PR #26, opened 2014-11-24
Julia.__init__ points directly to julia executable

mrocklin:
`jl_init_path` used to point to the julia directory and then we found the executable by looking into `./usr/bin/julia`.  This works in master but not 3.3.  What is the appropriate assumption here?  Assuming that it's not universally consistent perhaps it's best to specify the `julia` binary directly.


Keno:
This is handled now by asking Julia for it's HOME.

PR #33, opened 2015-02-13
make 'pyinitialize' work under windows

stonebig:


stonebig:
replace
https://github.com/JuliaLang/pyjulia/issues/23

julia/core.py
```diff
@@ -286,7 +289,11 @@ def __init__(self, init_julia=True, jl_init_path=None):
                     "Install PyCall by running the following line:\n"
                     """\tjulia -e 'Pkg.add("PyCall"); Pkg.update()'\n""")
             try:
-                self.call('pyinitialize(C_NULL)')
+                if os.name == "nt":
+                    # on windows C_NULL doesn't mean anything
+                    self.call('pyinitialize()')
```
jakebolewski: @stevengj is this true?

stonebig: more precisely, a variable -1 is asked to `Py_GetVersion` , which is clueless about its meaning.
I may be wrong on the true cause, but the fix works

stevengj: This doesn't seem right to me.  Running `pyinitialize()` tries to find `libpython` by running the `python` executable.  If you are initializing PyCall from *within* Python, you want it to use the `libpython` that is already loaded into memory.

If you pass `C_NULL` to `pyinitialize`, it initializes `libpython` [by special code](https://github.com/stevengj/PyCall.jl/blob/183c915804ae470cb0a6acff123551dd27dde43d/src/pyinit.jl#L135) that [loads the base module](https://github.com/JuliaLang/julia/blob/233ebd85552345bc550d112480688680d7b0d8f5/src/dlload.c#L87-L97).

stonebig: Well, with `C_NULL` on winpython, I have this error:
`symbol could not be found Py_GetVersion (-1): no error`


julia_c_null


Wuthout `C_NULL`, it works

stevengj: @stonebig, the problem here is that using `pyinitialize()` is extremely fragile in this context: if it loads a different `libpython` than the one that is currently running then everything will break.   We need to figure out what the underlying problem is.

stevengj: I think the problem is that `jl_load_dynamic_library(C_NULL)` on Windows is returning a reference to the running `libjulia` library (not what we want), whereas on Unix it returns `dlopen(NULL)` which is the handle for the main program (which is what we want).

stonebig: So, if my patch allows to have an "extremely fragile" pyjulia on windows, instead of an "already broken" one, shouldn't you call that "progress" and merge it ?

stevengj: I would rather have something that breaks noisily, and hence gets fixed, than something that breaks mysteriously when you do something unrelated on your system (e.g. install another `python` executable version in your path).

stevengj: In particular, it looks like the solution might be to have `julia/src/dlload.c` use `GetModuleHandle(NULL)` instead of `GetModuleHandleExW(...)` in `jl_load_dynamic_library_`; this will get the handle of the module for the running executable, which I think is what we need to look up symbols in a globally-loaded `libpython`, rather than the `libjulia` module handle.  (i.e. it seems to be the analogue of the `dlopen(NULL)` that `jl_load_dynamic_library_` is already doing on Mac and Linux systems).

@tkelman, @vtjnash, can you comment?

tkelman: No idea, sorry. I don't know any better here than to look through those MSDN pages.

stevengj:
I've just pushed a PyCall patch (stevengj/PyCall.jl@b4974b0fa209166a9ec078a066cf810760eeb8ff) that attempt to fix `pyinitialize(C_NULL)` on Windows.   @stonebig, can you try it?

julia/core.py
stevengj: On Windows, it looks like `GetModuleHandle(NULL)` is already called by Julia and stashed in `jl_exe_handle`.  I've pushed a patch to PyCall that uses this, and I'm hoping it fixes the problem.

stonebig:
Yes, I hope within the coming hour

stonebig:
ok, I did modify the source, I hope it recompiles automatically.

from within IDLE, result is, much quicker than before:
````
`>>> j = julia.Julia()
  File "D:\WinPython\basedir34\buildFlavorJuliab4\winpython-3.4.2.amd64\python-3.4.2.amd64\lib\site-packages\julia\core.py", line 294, in __init__
    self.call('pyinitialize(C_NULL)') # stonebig
  File "D:\WinPython\basedir34\buildFlavorJuliab4\winpython-3.4.2.amd64\python-3.4.2.amd64\lib\site-packages\julia\core.py", line 336, in call
    .format(exception_type, exception_msg, src))
julia.core.JuliaError: Exception 'UndefRefError' ocurred while calling julia code:
<couldn't get stack>

Code:
pyinitialize(C_NULL)

During handling of the above exception, another exception occurred:

  File "<pyshell#2>", line 1, in <module>
    j = julia.Julia()
  File "D:\WinPython\basedir34\buildFlavorJuliab4\winpython-3.4.2.amd64\python-3.4.2.amd64\lib\site-packages\julia\core.py", line 298, in __init__
    raise JuliaError("Failed to initialize PyCall package")
julia.core.JuliaError: Failed to initialize PyCall package
>>> 
````

Maybe I didn't test right.

stonebig:
I found back the fibonacci example and checked it works with my initial patch.


http://nbviewer.ipython.org/github/winpython/winpython_afterdoc/blob/master/docs/FlavorJulia/Using%20python%20julia%20module.ipynb

Keno:
I tested Windows as part of #54 and fixed any issues I encountered. Please try it out and file a new issue if there are any problems.

PR #43, opened 2015-11-09
Update core.py

nabusman:
pyinitialize() is deprecated in PyCall


Keno:
Current master does not call this function anymore.

Issue #38, opened 2015-04-13
v0.4 support


jpfairbanks:
I was trying to use this in an ipython notebook to support julia v0.4 but ran into this exception. 
`JuliaError: error starting up the Julia process.` The problem is in [core.py:237](https://github.com/JuliaLang/pyjulia/blob/master/julia/core.py#L237)
in 0.4 `Sys.dlpath` does not exist it was moved to `Libdl.dlpath`.

I don't know how large of an effort will be needed to get this working on 0.4 but I would rather file the issue than lose the information.

joschu:
I fixed the problem mentioned by jpfairbanks, but there's another problem that looks more difficult to fix, whenever you try to run `julia.Julia()`--`self.api.jl_call1` returns `None`:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "julia/core.py", line 304, in __init__
    for name, func in iteritems(base_functions(self)):
  File "julia/core.py", line 165, in base_functions
    names = julia.eval("names(Base)")
  File "julia/core.py", line 362, in eval
    "{}".format(src))
julia.core.JuliaError: ErrorException in Julia
```

leonardt:
I threw this snippet at the end of `eval` (same code used in `call`)
```
        exoc = self.api.jl_exception_occurred()
        if not res and exoc:
            exception_type = self.api.jl_typeof_str(exoc).decode('utf-8')
            try:
                exception_msg = self._capture_showerror_for_last_julia_exception()
            except UnicodeDecodeError:
                exception_msg = "<couldn't get stack>"
            raise JuliaError(u'Exception \'{}\' in Julia PyObject:\n{}\n\nCode:\n{}'
                             .format(exception_type, exception_msg, src))
```

and the traceback gives me a `MemoryError` which seems rather odd.

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/home/truongle/pyjulia/julia/core.py", line 300, in __init__
    for name, func in iteritems(base_functions(self)):
  File "/mnt/home/truongle/pyjulia/julia/core.py", line 165, in base_functions
    names = julia.eval("names(Base)")
  File "/mnt/home/truongle/pyjulia/julia/core.py", line 362, in eval
    .format(exception_type, exception_msg, src))
julia.core.JuliaError: Exception 'MemoryError' in Julia PyObject:
<couldn't get stack>

Code:
names(Base)
```

leonardt:
[...]

PR #44, opened 2015-11-20
WIP - buggy Julia 0.4 support

benmoran:
I've been trying to make the modifications discussed in issue #38.  It's not there yet - don't merge it!

I can eval quite a few Julia commands, but still get errors and segfaults (running on Ubuntu 14.04, system Python 2.7.6 and Julia 0.4.1 from julia-releases).

The main issue I'm fighting at the moment is that ``import julia; j = julia.Julia()`` gives:

    InitError(:PyCall,PyError (:PyCapsule_Import) <type 'exceptions.ImportError'>
    ImportError('PyCapsule_Import could not import module "datetime.datetime_CAPI"',)

It looks similar to the bug reported here but not resolved: https://github.com/stevengj/PyCall.jl/issues/178

Rebuilding PyCall doesn't seem to help.

I wonder if it is because the datetime module has already been initialized?

### DIFF
--- a/julia/core.py
+++ b/julia/core.py
@@ -20,7 +20,6 @@ import os
 import sys
 import keyword
 import subprocess
-import warnings
 
 from ctypes import c_void_p as void_p
 from ctypes import c_char_p as char_p
@@ -250,20 +249,14 @@ class Julia(object):
                      println(Libdl.dlpath(Libdl.dlopen(\"libjulia\")))
                      """])
                 JULIA_HOME, libjulia_path = juliainfo.decode("utf-8").rstrip().split("\n")
-                libjulia_dir = os.path.dirname(libjulia_path)
             except:
                 raise JuliaError('error starting up the Julia process')
 
             if not os.path.exists(libjulia_path):
                 raise JuliaError("Julia library (\"libjulia\") not found! {}".format(libjulia_path))
             self.api = ctypes.PyDLL(libjulia_path, ctypes.RTLD_GLOBAL)
-            self.api.jl_init.arg_types = [char_p, char_p]
-            sysimg = self.find_sysimg(libjulia_dir)
-            path = os.path.dirname(sysimg)
-            sysfile = os.path.basename(sysimg)
-            #print "Using path %s, sysfile %s" % (path, sysfile)
-            self.api.jl_init_with_image(path.encode("utf-8"),
-                                        sysfile.encode("utf-8"))
+            self.api.julia_init.argtypes = [ctypes.c_int]
+            self.api.julia_init(1)
 
         else:
             # we're assuming here we're fully inside a running Julia process,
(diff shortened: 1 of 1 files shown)


stevengj:
@benmoran, thanks for working on this.  Do you get that error from `using PyCall` in Julia?

benmoran:
No, it's fine if I use PyCall in Julia. In fact I can run Julia, use PyCall, @pyimport julia and then observe the error in the same Julia process where it had previously worked.

stevengj:
What happens if you do:
```jl
using PyCall
import PyCall: @pycheckn, @pysym, PyDateTime_CAPI
unsafe_load(@pycheckn ccall((@pysym :PyCapsule_Import),
                                            Ptr{PyDateTime_CAPI}, (Ptr{UInt8}, Cint),
                                            "datetime.datetime_CAPI", 0))
```
in Julia?  Calling this multiple times works for me, so I don't think it's a problem with it already being loaded.

sebastroy:
I have been working on this sporadically as well. I use python 3.5 though and I get different errors. In my case I can't manage to get the 

         res = self.api.jl_call2(void_p(self.api.convert), void_p(self.api.PyObject),void_p(ans))

working. I tried a handful of variation on the "self.api.convert" generation but they all generate a segfault when the one quoted above is invoked. If I skip the "base" part I can get it working partially. Partially in the sense that I can get julia to execute commands appropriately but I can't get the results back as a python object.

benmoran:
@stevengj Calling unsafe_load ... as above works fine in Julia for me as well.

sebastroy:
Allright I got it I think! I can build without errors, import julia and declare a j = julia.Julia() without error. I will see if I can make it really work and submit a patch. Way less radical changes than this pull request BTW. But then again I haven't studied the above changes. My gut feeling is that it's not necessary to make so many changes but maybe I'm missing something. Also, not sure about python 2 vs 3 compatiblity...

benmoran:
@sebastroy you're right, this pull request also has other changes to try to make Julia stack traces better visible - I was finding this helpful for debugging.

benmoran:
There are two things going on outside PyJulia - the Julia version and PyCall version.  I have a working setup of older code, and a new one.
 
 - working: (Julia 0.3, pyjulia master, PyCall 32b10a5...)  
 - not working: (Julia 0.4.1, benmoran/pyjulia julia-0.4, PyCall master) 

The second one gives the ``PyCapsule_Import could not import module "datetime.datetime_CAPI"`` error on ``Julia.call("using PyCall")``.

This might be due to changes in PyCall rather than just the Julia version number.  (I can't check them separately because I can't run the new PyCall with Julia 0.3 or vice versa, due to dependencies being incompatible.)

The main change in the new PyCall is that the ``PyCall.pyinitalize()`` function has gone away.  I removed the call entirely in the pyjulia branch here - as far as I could tell it's no longer necessary.  But if I try to ``api.call("@pyimport datetime")`` in my working Julia 0.3 setup after ``using PyCall`` but before ``pyinitialize(C_NULL)`` then it gives me the same error from PyCapsule_Import.  (In the Julia 0.3 shell itself ``using PyCall; @pyimport datetime`` works fine.)

So @stevengj - am I missing some necessary initialization step for the new version of PyCall?

(Also, as far as I can tell, the ImportError results not because the datetime .so file fails to dlload correctly, but because "datetime" doesn't appear in sys.modules afterwards).

benmoran:
Superceded by https://github.com/JuliaLang/pyjulia/pull/47

PR #46, opened 2015-11-24
Changes in julia/core.py related to pull request #44 and issue #38

sebastroy:
Changed Sys to Libdl
        Changed sys.jl to sys.so
        pyinitialize is removed

        Replace jl_call1 to jl_call2 and use convert
        to return a PyObject. see JuliaLang/pyjulia#38

### DIFF
--- a/julia/core.py
+++ b/julia/core.py
@@ -26,6 +26,7 @@ from ctypes import c_void_p as void_p
 from ctypes import c_char_p as char_p
 from ctypes import py_object
 
+
 # this is python 3.3 specific
 from types import ModuleType, FunctionType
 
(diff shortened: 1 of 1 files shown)


sebastroy:
Ok still some work to do... The native "base" methods works but the basic "call" doesn't and I'm guessing the convert thing should happen there too...

sebastroy:
here is an example of the behavior I have in the python interpreter:

  >>> j.sum([1,2])
  3
  >>> j.call('sum([1,2])') 
  RuntimeError: Julia exception: MethodError(call,("sum([1,2])",))
  >>>

Keno:
Most of these changes have made it in. If you would like to update expression printing, please submit a separate pull request.

Issue #51, opened 2016-02-12
Custom builds of PyCall.jl


benmoran:
It's currently possible to get pyjulia to work, at least on Linux, by using PyCall.jl at tag v1.3.0 together with this PR for PyJulia: https://github.com/JuliaLang/pyjulia/pull/47

But a proper solution is still needed.  @stevengj has some suggestions here:
https://github.com/stevengj/PyCall.jl/pull/219#issuecomment-180543881

In particular he suggests that each pyjulia ought to make a separate Pkg.clone of PyCall, and Pkg.build it specializing to the instance of Python that is being used.

stevengj:
This should probably be combined with an PR to PyCall, so that if you set some environment variable during the `Pkg.build` phase then it looks up CPython symbols in the global namespace (for cases where `libpython` is statically linked into `python`).  In this case, I'm hoping you can just use `ccall(:Py_SomeSymbol, ...)`, which should be a trivial change to the `@pysym` macro in PyCall.

stevengj:
I would also suggest cloning into a "private" directory for `pyjulia`, and `push!(mydirectory, LOAD_PATH)` so that Julia can find it.  You can also push a private directory to `Base.LOAD_CACHE_PATH` so that the `PyCall.ji` cache file will be stored there.   In particular, do something like:
```jl
using Compat
push!(mydirectory, LOAD_PATH)
push!(mydirectory, Base.LOAD_CACHE_PATH)
using PyCall
pop!(LOAD_PATH)
pop!(Base.LOAD_CACHE_PATH)
```
when you are initializing `pyjulia`.   (Do `using Compat` first so that if it has been updated and needs to be recompiled, it's cache file will go into the standard Julia cache path rather than into your private cache path.)

Keno:
The real problem here is that is python the precompile environment needs to look the same as the runtime environment. In particular that means that in the general case python needs to be in charge of the precompile environment as well. Talking with @vtjnash, we could create a python script that takes the appropriate arguments and pretends to be julia closely enough for julia to use it for precompilation.

stevengj:
@Keno, I'm not sure I follow you.   As long as we build with the right version of Julia, and use the pyjulia `Pkg.build("PyCall")` output, I don't see why precompilation within `julia` shouldn't work.

Keno:
Because that will try to load the symbols from the executable environment, which won't be there in the precompile process.

Keno:
[...]

PR #54, opened 2016-07-07
WIP: pyjulia updates

Keno:
This does two things:

1. A rebase of some of the previous pull requests to bring pyjulia up to date with more modern versions of julia
2. Attempt to solve the challenges encountered with PyCall (i.e. #51)

The approach taken to 2) is to

a. Separate out cache files into a separate directory to reflect the fact that this is a quite different environment. It may be theoretically possible to share packages that do not depend on PyCall, but this could turn out to be quite brittle and the necessary logic does not currently exist in julia to handle this situation. I think a clean separation as is done here is the best way to go.
b. Use a python script that calls the argument parsing routines directly. This is necessary because the precompilation and runtime environment needs to be the same in order for PyCall to be able to be used without major modification (small PyCall PR to enable loading in-process symbols is forthcoming). I played with a couple of variants here, but I think I've arrived at a very simple/effective solution.

Remaining todo:
- [x] Only enable this if python is statically linked?
- [x] COFF Wrapper for windows
- [x] Test on other Unix platforms
- [x] Mac CI
- [x] Enable AppVeyor
- [x] Make sure new code path is actually tested
- [x] Windows cross-version test on CI
- [x] Update COFF wrappers to 32bit
- [x] https://github.com/JuliaLang/julia/pull/17458 to fix tests


julia/core.py
```diff
@@ -297,6 +297,14 @@ def __init__(self, init_julia=True, jl_runtime_path=None, jl_init_path=None,
         self.api.jl_exception_clear()
 
         if init_julia:
+            # Replace the cache directory with a private one. PyCall needs a different
+            # configuration and so do any packages that depend on it. Ideally, we could
```
tkelman: why can't we use the normal pycall?

Keno: Needs a different configuration (in-process vs shared library).

julia/core.py
```diff
@@ -271,42 +280,66 @@ def __init__(self, init_julia=True, jl_init_path=None):
...
+            os.environ["PYCALL_PYTHON_EXE"] = sys.executable
+            os.environ["PYCALL_JULIA_HOME"] = JULIA_HOME
+            self._call(u"eval(Base,:(JULIA_HOME=\"/Users/kfischer/Projects/pyjulia/fake-julia\"))")
+            # Replace the cache directory with a private one. PyCall needs a different
+            # configuration and so do any packages that depend on it. Ideally, we could
+            # detect packages that depend on PyCall and only use LOAD_CACHE_PATH for them
+            # but that would be significantly more complicated and brittle, and may not
+            # be worth it.
+            self._call(u"empty!(Base.LOAD_CACHE_PATH)")
+            self._call(u"push!(Base.LOAD_CACHE_PATH, abspath(Pkg.Dir._pkgroot()," +
```
tkelman: using internal methods here doesn't look like a great idea for stability's sake

Keno: You mean `_pkgroot`? Unfortunately that's how to determine where this directory is. If you want to advocate for exporting that method from Base, be my guest.

tkelman: Or use the existing `LOAD_CACHE_PATH` and just modify that.

Keno: Seems like an ok option/

julia/core.py
```diff
@@ -193,6 +193,15 @@ def module_functions(julia, module):
             pass
     return bases
 
+def determine_if_statically_linked():
+    """Determines if this python executable is statically linked"""
+    # Windows and OS X are generally always dynamically linked
+    if not sys.platform.startswith('linux'):
+        return False
+    lddoutput = subprocess.check_output(["ldd",sys.executable])
+    return not lddoutput.conains("libpython")
```
tkelman: typo?

Keno: Yes already fixed locally. Give me a few more mins and I'll push all the windows code.

Keno:
There's a travis failure here I don't understand: `Type str doesn't support the buffer API`. Works fine for me locally, even with python 3.

appveyor.yml
```diff
@@ -28,10 +28,6 @@ environment:
   - JULIAVERSION: "julialang/bin/winnt/x64/0.4/julia-0.4-latest-win64.exe"
     PYTHONDIR: "C:\\Python35-x64"
 
-branches:
-  only:
-    - clean-julia-0.4
```
tkelman: should probably change this to only master, otherwise you get duplicate builds (branch and pr)

.travis.yml
```diff
@@ -1,14 +1,24 @@
...
 notifications:
   email: false
 before_install:
   - sudo add-apt-repository ppa:staticfloat/julia-deps -y
-  - sudo add-apt-repository ppa:staticfloat/juliareleases -y
+  if [ "$JULIA_VERSION" = "LAST_RELEASE" ]; then
+      - sudo add-apt-repository ppa:staticfloat/juliareleases -y
+  else
+      - sudo add-apt-repository ppa:staticfloat/juliareleases -y
+  end
```
tkelman: We shouldn't be using the ppa here, it's effectively deprecated. Also doesn't help on osx.

Keno: What's the suggested way?

tkelman: could modernize https://github.com/JuliaLang/julialang.github.com/blob/master/install-julia.sh (and take out the ppa parts) which approximates what `language: julia` does on travis. Travis doesn't support multiple language settings afaik, and `language: python` is probably more complicated.

stevengj:
Do we really want `julia-debug.exe` to be checked into git?

Keno:
Yes, it's not the real julia-debug.exe, it's just a very small COFF wrapper.

tkelman:
win32 or win64?

Keno:
win64, though as long as your os is 64bit it doesn't matter. I suppose I could just compile it as 32bit and have it work anywhere.

tkelman:
Your Travis syntax is broken here and is eating up a lot of JuliaLang's queue. Could you enable it on your fork and work things out on a different branch for now?

tkelman:
We could also move this package to JuliaInterop.

Keno:
Yep, will move to JuliaInterop

ararslan:
I enabled AppVeyor here so if you have pyjulia under your AppVeyor account, you can modify it to point here. :)

.travis.yml
```diff
@@ -46,6 +46,7 @@ matrix:
 notifications:
   email: false
 before_script:
+  - echo ./ci/install-julia.sh $JULIA_VERSION
```
tkelman: if you end up needing to make changes to this script I'd still rather "upstream" them and have just one copy that lives in the website repo, once you get it all working

appveyor.yml
```diff
@@ -0,0 +1,60 @@
...
+install:
+# Download most recent Julia Windows binary
+  - ps: (new-object net.webclient).DownloadFile(
+        $("http://s3.amazonaws.com/"+$env:JULIAVERSION),
+        "C:\projects\julia-binary.exe")
+# Run installer silently, output to C:\projects\julia
+  - C:\projects\julia-binary.exe /S /D=C:\projects\julia
+
+build_script:
+  - "SET PATH=%PYTHONDIR%;%PYTHONDIR%\\Scripts;%PATH%"
```
tkelman: add the fast fail snippet for repeated pushes to the same pr here

Keno: What's the fast fail snippet?

tkelman: it's a powershell web request thing that should be in Example.jl's appveyor.yml

Keno:
Windows test failure is due to it trying to find `rint` in the executable and of course not finding it since python is not linked against openlibm. I'm not sure why that's happening, but since it seems fixed on 0.5, I'm not going to put too much effort into attempting to figure out why.

fake-julia/julia.py
```diff
@@ -0,0 +1,74 @@
...
+jl_base_module = c_void_p.in_dll(libjulia, "jl_base_module")
+_start = libjulia.jl_get_global(jl_base_module, libjulia.jl_symbol(b"_start"))
+args = (c_void_p * 1)()
+args[0] = _start
+libjulia.jl_apply_generic(args, 1)
+libjulia.jl_atexit_hook(0)
+
+# As an optimization, share precompiled packages with the main cache directory
+libjulia.jl_eval_string(b"""
+    outputji = Base.JLOptions().outputji
```
tkelman: should this use a `let` to avoid always defining this?

Keno: This is after the incremental image has been written, so it'll go away.

test/test_core.py
```diff
@@ -44,11 +45,9 @@ def add(a, b):
             return a + b
         self.assertEqual([1, 4, 9],
                          list(julia.map(lambda x: x * x, [1, 2, 3])))
-        print(julia.map(lambda x: x + 1,
-                                   array.array('I', [10, 10, 10])))
-        self.assertEqual([11, 11, 11],
+        self.assertTrue(np.array_qual([11, 11, 11],
                          julia.map(lambda x: x + 1,
-                                   array.array('I', [10, 10, 10])))
+                                   array.array('I', [10, 10, 10]))))
```
ararslan: I can only imagine how Lisp programmers must feel all the time

Keno:
Ok, I'm gonna go and merge this. Travis OS X is complaining, but that's an ecosystem wide problem right now, so oh well. If it turns out to be broken after, I'll fix that separately.

PR #82, opened 2017-02-27
fix issue with python >= 2.7.13 and < 3

TsurHerman:
see issue: https://bugs.python.org/issue29082

this PR fixes that for pyjulia

### DIFF
--- a/julia/core.py
+++ b/julia/core.py
@@ -274,6 +274,8 @@ class Julia(object):
             self._debug("JULIA_HOME = %s,  libjulia_path = %s" % (JULIA_HOME, libjulia_path))
             if not os.path.exists(libjulia_path):
                 raise JuliaError("Julia library (\"libjulia\") not found! {}".format(libjulia_path))
+            if sys.version_info >= (2,7,13) and sys.version_info < (3,0,0):
+                libjulia_path = libjulia_path.encode("ascii")
             self.api = ctypes.PyDLL(libjulia_path, ctypes.RTLD_GLOBAL)
             if not jl_init_path:
                 if jl_runtime_path:
--- a/julia/core.py
+++ b/julia/core.py
@@ -274,8 +274,13 @@ class Julia(object):
             self._debug("JULIA_HOME = %s,  libjulia_path = %s" % (JULIA_HOME, libjulia_path))
             if not os.path.exists(libjulia_path):
                 raise JuliaError("Julia library (\"libjulia\") not found! {}".format(libjulia_path))
-            if sys.version_info >= (2,7,13) and sys.version_info < (3,0,0):
+                
+            # fixes a specific issue with python 2.7.13 
+            # ctypes.windll.LoadLibrary refuses unicode argument
+            # http://bugs.python.org/issue29294
+            if sys.version_info == (2,7,13):                   
                 libjulia_path = libjulia_path.encode("ascii")
+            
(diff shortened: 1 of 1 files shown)


TsurHerman:
Any comments guys?

stevengj:
This doesn't seem right... what happens if the path contains non-ASCII characters?

TsurHerman:
It seems that the common belief in the net is that it will be fixed in 2.7.14 .. so I don't know what to say.
you can delete this PR if you are not going to merge it . it is a bug that will go away in the future.

stevengj:
So in Python 2.7.13 there is no way to load a DLL from a path that contains non-ASCII characters?

TsurHerman:
it seems so
http://bugs.python.org/issue29294

julia/core.py
```diff
@@ -274,6 +274,8 @@ def __init__(self, init_julia=True, jl_runtime_path=None, jl_init_path=None,
             self._debug("JULIA_HOME = %s,  libjulia_path = %s" % (JULIA_HOME, libjulia_path))
             if not os.path.exists(libjulia_path):
                 raise JuliaError("Julia library (\"libjulia\") not found! {}".format(libjulia_path))
+            if sys.version_info >= (2,7,13) and sys.version_info < (3,0,0):
```
stevengj: So this should be `< (2,7,14)`?

stevengj: Or better yet, just `== (2,7,13)`?

stevengj: And add a comment linking to http://bugs.python.org/issue29294

stevengj:
LGTM

stevengj:
Why are we checking ≤ 2.1.14, if it will be fixed in that release?  Why not == 2.7.13?

TsurHerman:
because I tested it today on a new system and:

```
In [1]: import sys

In [2]: sys.version_info == (2,7,13)
Out[2]: False

In [3]: sys.version_info <= (2,7,13)
Out[3]: False

In [4]: sys.version_info >= (2,7,13)
Out[4]: True

In [5]: sys.version_info >= (2,7,13) and  sys.version_info <= (2,7,14)
Out[5]: True

In [6]: sys.version_info
Out[6]: sys.version_info(major=2, minor=7, micro=13, releaselevel='final', serial=0)
```

stevengj:
That's because `sys.version_info` includes a "micro" version number).  You can do:
```
sys.hexversion & 4294967040 == 34016512
```
to check for 2.7.13 (where the bitmask strips off the micro version to only leave major/minor/patch).

TsurHerman:
I find that the meaning of the operation is lost in this way.. and that it requires knowledge of the inner representation of version info, which I do not want to depend upon.
feel free to change it to a format you like , hopefully this issue will go away soon

stevengj:
`sys.version_info >= (2,7,13) and  sys.version_info <= (2,7,14)` is wrong if it is fixed in 2.7.14.   Maybe you meant `sys.version_info >= (2,7,13) and  sys.version_info < (2,7,14)`?

TsurHerman:
O.k , yes I meant it to be up-to (2,7,14) but got lazy removing the extra equal sign, reasoning that anyway 
version (2,7,14,0,0) only exists in theory when the new release is planned and by now it is already (2,7,14, and something) so the intended behaviour will be satisfied.

... tl;dr
I am removing the unnecessary equal sign

stevengj:
LGTM. The Travis failures with 0.6 look unrelated

PR #103, opened 2017-06-13
WIP: fixes for 0.7

ihnorton:
Putting this here for someone (maybe future me).

~@klickverbot the self.api.show definition needs to be moved earlier in the file so it is available if an error is thrown during initialization. (I'll remove those commits before merging).~

also, FYI: pyjulia is broken on 0.6 and nightly on Windows for now, see https://github.com/JuliaLang/julia/issues/22320.


Keno:
@ihnorton Any objections to me merging this with some additional fixes for 0.7?

Keno:
Done in #110.

PR #107, opened 2017-07-19
Python 3.5 is supported and probably 3.6

PallHaraldsson:
3.4 was tested, now only 3.5. Add more tests?


mmagnuski:
pyjulia has build errors (even before your PR) on Travis - I think these should be fixed before expanding CI tests to other python versions.

musm:
bump

PR #111, opened 2017-07-28
Drop 0.4 from AppVeyor, add 0.5/0.6

Keno:


Keno:
Fails on 0.6, but that's known (https://github.com/JuliaLang/julia/issues/22866)

PR #112, opened 2017-07-28
Increase python stack size on CI

Keno:
This is the current recommended workaround, so we might as well make sure it works.


Keno:
Well, didn't quite fix all of the 0.6 issues, but good enough for now.

Issue #118, opened 2017-08-31
Make pyjulia easy to install


madeleineudell:
I'd like to make it easy for Python denizens to use my Julia packages, so am writing a thin Python wrapper using pyjulia. But to make my wrapper easily installable, I'd need pyjulia to be easily installable, too. Would it be possible to allow installation via pip/conda/etc?

This would move us in the direction of having pyjulia be as easy to use from Python as PyCall is from Julia.

Bonus points if it auto-installs Julia upon not finding a valid Julia installation.

ChrisRackauckas:
>Bonus points if it auto-installs Julia upon not finding a valid Julia installation.

Up for this. @tkf what would it take to achieve this? I'm curious how possible it is.

tkf:
In principle, we can "just" write a simile installer by porting the existing shell script https://github.com/JuliaCI/install-julia to Python.  But dealing with integrity check and security concerns (like checking PGP signatures) is a very daunting task.  It essentially means that PyJulia becomes a package manager which I'd like to avoid.  But maybe adding `unsafe=True` option to `julia.install()` function https://pyjulia.readthedocs.io/en/latest/installation.html#step-3-install-julia-packages-required-by-pyjulia that auto-installs everything is an OK compromise.

ChrisRackauckas:
@christopher-dG let's discuss how this vednering of Julia might be possible. Maybe just detect the architecture and install the generic binary? Maybe something from binary builder can be helpful. @staticfloat might be helpful as well.

christopher-dG:
So I think the best option is to enhance the `pyjulia.install()` function as @tkf mentioned rather than putting Julia installation to install or load time. We could use Julia_jll, or just download the tarballs from julialang.org.

I guess we should do nothing if there is already a Julia v1+ installed, and install to somewhere in the pyjulia installation otherwise.

christopher-dG:
I wrote some ugly code that downloads, verifies (hash + GPG), and extracts Julia, obviously it needs a lot less stuff hard-coded and probably some Python 2 compat work but this is mostly just to show that the verification is not actually that difficult.

```py
def install_julia():
    # Check if Julia is already installed
    if which("julia") is not None:
        return "julia"
    dest = os.path.expanduser("~/.local/share/pyjulia/installs/julia-1.4.2")
    os.makedirs(os.path.dirname(dest), exist_ok=True)
    julia_bin = os.path.join(dest, "bin", "julia")
    if os.path.isdir(dest):
[...]

PR #120, opened 2017-09-06
Update release instructions and bump to 0.1.4

Carreau:
Hi all, 

The author of https://github.com/JuliaPy/pyjulia/issues/118 (@madeleineudell) contacted me asking if I can release – as I am the one with creds on PyPI. As usual, happy to also give access to anyone else admin on this repo requesting for it. 

So here is the head of what I'm going to release if this is ok. It would be nice to merge it and tag it for consistency (I can't, dont' have the rights on the repo).

If this get no answer I'll publish anyway in a couple of days to make it easy for contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,15 @@ Your python installation must be able to call Julia.  If your installer
 does not add the Julia binary directory to your `PATH`, you will have to
 add it.
 
-Then finally you have to install pyjulia. You may clone it directly to your home directory.
+Then finally you have to install pyjulia. 
+
+To get released versions you can use:
+
+```
+pip install pyjulia
+```
+
+You may clone it directly to your home directory.
 
 ```
 git clone https://github.com/JuliaPy/pyjulia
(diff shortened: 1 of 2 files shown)


madeleineudell:
@Carreau , thanks!

May I ask your view on whether pip installing pyjulia ought ever to install julia itself? As a package developer, I'd like to be able to have users install my python package without needing to install julia manually (or even, for them to think about the fact that the backend is in julia).

Carreau:
I would suggest reading [this blog post](http://www.curiousefficiency.org/posts/2016/09/python-packaging-ecosystem.html) from Nick Coghlan (core CPython dev). 

Having pip install other programming languages (or shared binaries) is a complex problem, and this is better suited for distribution and package manager of distributions (like conda, if you are a datascientist, apt if you are a linux guru, ... ).  You can try to bend pip in order to install Julia but there will certainly be a lot of pains if you try. 

On the jupyter side we tried to happer pip into install js package from npm, and it was not a great story; So I would advise against it at least for now.

Carreau:
Uploaded to PyPI:
https://pypi.python.org/pypi/julia

stevengj:
See also https://github.com/tkf/conda-julia .... would be nice to have `conda add pyjulia`.

Carreau:
We should be able to request a pyjulia-feedstock on conda-forge.

PR #127, opened 2017-10-06
Corrected the pip install command

abalkin:
The latest version is published on PyPI as "julia".  See <https://pypi.python.org/pypi/julia>.


Keno:
Picked up as #143, sorry for missing this.

Issue #135, opened 2017-11-28
pip install julia?


zsunberg:
I have installed pyjulia on a couple machines, but have never gotten
```
pip install pyjulia
```
to work. Instead I found
```
pip install julia
```
to work. Is that the correct command? Should the README be updated to reflect that?

Naereen:
The README *should* be updated! Anyone in the dev team agrees?

I will write the PR as soon as anyone confirm.

terasakisatoshi:
I confirmed `pip install julia` works

Naereen:
OK. I opened the PR, if any dev can merge it #143.

Naereen:
@Keno OK :+1:

PR #143, opened 2018-01-15
pip install julia and not pyjulia

Naereen:
Cf. #135


Naereen:
Something is wrong here, `pip install pyjulia` don't work but the ci builds both failed.

PR #149, opened 2018-03-24
Modernize install-julia.sh for CI and drop Julia v0.5

rdeits:
I noticed that many of the Travis builds were failing, and I suspect it might be because the `install-julia.sh` script is using outdated Julia download URLs. This should update the download URLs to their new homes. 

I also noticed that this script accepted `download/win32` etc. as arguments, but (a) their URLs were also out of date and (b) if those URLs existed, the script would have done nothing with them on windows and tried to feed a windows `.exe` directly into `tar` on linux. So I've just deleted those arguments. Please let me know if you want them back.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,3 +72,4 @@ script:
   - python --version
   - if [ "$CROSS_VERSION" = "1" ]; then /usr/bin/python -m unittest discover; fi
   - python -m unittest discover
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ matrix:
           - PYTHON_VERSION=py3
           - JULIA_VERSION=juliareleases
         os: osx
+ allow_failures:
+      - env: JULIA_VERSION=julianightlies
 notifications:
   email: false
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ matrix:
           - PYTHON_VERSION=py3
           - JULIA_VERSION=juliareleases
         os: osx
- allow_failures:
+  allow_failures:
       - env: JULIA_VERSION=julianightlies
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,15 @@ matrix:
         os: osx
   allow_failures:
       - env: JULIA_VERSION=julianightlies
+      - env:
+          - JULIA_VERSION=julianightlies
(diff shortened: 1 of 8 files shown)


rdeits:
I've also dropped Julia v0.5 on appveyor since it's no longer supported by PyCall.jl

rdeits:
Ok, I've made some progress on this. I had to mark the v0.7 tests as allowed failures, since PyCall doesn't work with v0.7. It looks like there's also an existing problem with some of the v0.6 builds on windows: https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.45/job/4eg7o7bh7fh7t6cu#L159 which existed previously: https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.35/job/d6gu3c898kma3jid#L95

tkf:
I added a one-line fix in #161 for appveyor. @rdeits Can you pull https://github.com/JuliaPy/pyjulia/pull/161/commits/20e3c717ceab7ffcafaf4d6f93e8d3a8442d71de ? (BTW, thanks for all the fixes!)

Or the maintainers can pull #161 so that this PR would be automatically closed as well.  Either way, I'm happy as long as the CI is up :)

rdeits:
@tkf thanks! I've pulled in https://github.com/JuliaPy/pyjulia/commit/20e3c717ceab7ffcafaf4d6f93e8d3a8442d71de

rdeits:
Ok, all checks are passing. Good to go?

tkf:
In osx build with `PYTHON_VERSION=py3`, it looks like Python 2 is used instead of 3: https://travis-ci.org/JuliaPy/pyjulia/jobs/373733201#L256

tkf:
I tried to fix it in #163. But it turned out `python3` is not installed with current `.travis.yml`:
https://travis-ci.org/JuliaPy/pyjulia/jobs/380045548#L262

Actually, `brew list python &>/dev/null || brew upgrade python` outputs nothing:
https://travis-ci.org/JuliaPy/pyjulia/jobs/380045548#L79
...so it means `brew` in travis cannot install Python 3?

I'm not familiar with `brew` so I think I'll not try to fix it further.

Anyway, I think this PR is better be merged, even without solving this problem.

tkf:
Actually I think now I solved it in #163!

rdeits:
Great! I'll close this in favor of #163

PR #151, opened 2018-04-14
Fix exception printing

fperez:
The Julia exception objects don't have a `.message` attribute, we can just print the plain object.


tkf:
Merged via #178. Thanks for the fix!

PR #160, opened 2018-04-28
Do not show exception in non-debugging mode

### LINKED ISSUE #159: Julia exceptions are unconditionally printed
I noticed that `import julia.<julia module>` prints a lot of `UndefVarError` when `<julia module>` exports undefined variables.  I think exporting undefined variables is better be avoided ideally.  However, it seems that pyjulia explicitly allows it by suppressing all exceptions

https://github.com/JuliaPy/pyjulia/blob/2eb704c33d6f639b41300023e16f910b5945600e/julia/core.py#L155-L162

Anyway, here is an example:

```pycon
Python 3.6.4 (default, Jan  5 2018, 02:35:40)
[GCC 7.2.1 20171224] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import julia
>>> j = julia.Julia()
>>> j.using('DiffEqBase')
>>> import julia.DiffEqBase
UndefVarError(:ConstantLagDDEProblem)UndefVarError(:ConstantLagDDEProblem)UndefVarError(:DAEParameterizedFunction)UndefVarError(:DAEParameterizedFunction)UndefVarError(:DDEParameterizedFunction)UndefVarError(:DDEParameterizedFunction)UndefVarError(:DEParameters)UndefVarError(:DEParameters)UndefVarError(:HasParamFuncs)UndefVarError(:HasParamFuncs)UndefVarError(:Mesh)UndefVarError(:Mesh)UndefVarError(:OutputParameterizedFunction)UndefVarError(:OutputParameterizedFunction)UndefVarError(:ParameterizedFunction)UndefVarError(:ParameterizedFunction)UndefVarError(:ProbParameterizedFunction)UndefVarError(:ProbParameterizedFunction)UndefVarError(:SecondOrderSDEProblem)UndefVarError(:SecondOrderSDEProblem)UndefVarError(:construct_correlated_noisefunc)UndefVarError(:construct_correlated_noisefunc)UndefVarError(:has_paramfuncs)UndefVarError(:has_paramfuncs)UndefVarError(:is_sparse_noise)UndefVarError(:is_sparse_noise)UndefVarError(:isautodifferentiable)UndefVarError(:isautodifferentiable)UndefVarError(:noise_class)UndefVarError(:noise_class)UndefVarError(:num_params)UndefVarError(:num_params)UndefVarError(:param_values)UndefVarError(:param_values)UndefVarError(:set_param_values!)UndefVarError(:set_param_values!)UndefVarError(:white_noise_func_wrapper)UndefVarError(:white_noise_func_wrapper)UndefVarError(:white_noise_func_wrapper!)UndefVarError(:white_noise_func_wrapper!)>>>
```

When I initialize it with `julia.Julia(debug=True)`, I get

```
...
exception occured? None
exception occured? None
exception occured? None
exception occured? 140476532807120
Retrieving exception infos...
libjulia stderr = 140476567600544
UndefVarError(:white_noise_func_wrapper)show called ...
exception occured? 140476532807296
Retrieving exception infos...
libjulia stderr = 140476567600544
UndefVarError(:white_noise_func_wrapper)show called ...
exception occured? 140476532807472
Retrieving exception infos...
libjulia stderr = 140476567600544
UndefVarError(:white_noise_func_wrapper!)show called ...
exception occured? 140476532807648
Retrieving exception infos...
libjulia stderr = 140476567600544
UndefVarError(:white_noise_func_wrapper!)show called ...
```

It looks like this happens because `Base.show` is called whenever exception occurred:

https://github.com/JuliaPy/pyjulia/blob/2eb704c33d6f639b41300023e16f910b5945600e/julia/core.py#L416-L432

But why do you need to print it when `JuliaError` is going to be raised anyway?

(see also: https://github.com/JuliaDiffEq/diffeqpy/pull/24)

tkf:
fixes #159

### DIFF
--- a/julia/core.py
+++ b/julia/core.py
@@ -428,11 +428,22 @@ class Julia(object):
             self._debug("show called ...")
         # self.api.jl_printf(self.api.jl_stderr_stream(), "\n");
 
-        res = self.api.jl_call2(void_p(self.api.convert), void_p(self.api.PyObject), void_p(exoc))
+        # If, theoretically, an exception happens in early stage of
+        # self.add_module_functions("Base"), showerror and sprint as
+        # below does not work.  Let's use jl_typeof_str in such case.
+        try:
+            sprint = self.sprint
+            showerror = self.showerror
+        except AttributeError:
+            res = None
+        else:
+            res = self.api.jl_call2(void_p(self.api.convert),
+                                    void_p(self.api.PyObject),
+                                    void_p(exoc))
         if res is None:
             exception = self.api.jl_typeof_str(exoc).decode('utf-8')
         else:
-            exception = self.sprint(self.showerror, self._as_pyobj(res))
+            exception = sprint(showerror, self._as_pyobj(res))
         raise JuliaError(u'Exception \'{}\' occurred while calling julia code:\n{}'
                          .format(exception, src))
 
--- a/julia/core.py
+++ b/julia/core.py
@@ -428,9 +428,13 @@ class Julia(object):
             self._debug("show called ...")
         # self.api.jl_printf(self.api.jl_stderr_stream(), "\n");
 
-        exception_type = self.api.jl_typeof_str(exoc).decode('utf-8')
+        res = self.api.jl_call2(void_p(self.api.convert), void_p(self.api.PyObject), void_p(exoc))
+        if res is None:
+            exception = self.api.jl_typeof_str(exoc).decode('utf-8')
+        else:
+            exception = self.sprint(self.showerror, self._as_pyobj(res))
         raise JuliaError(u'Exception \'{}\' occurred while calling julia code:\n{}'
-                         .format(exception_type, src))
+                         .format(exception, src))
 
     def _typeof_julia_exception_in_transit(self):
         exception = void_p.in_dll(self.api, 'jl_exception_in_transit')
@@ -453,7 +457,10 @@ class Julia(object):
         res = self.api.jl_call2(void_p(self.api.convert), void_p(self.api.PyObject), void_p(ans))
 
         if res is None:
-            self.check_exception("convert(PyCall.PyObject, %s)" % src)
+            self.check_exception(src)
+        return self._as_pyobj(res, "convert(PyCall.PyObject, {})".format(src))
+
+    def _as_pyobj(self, res, src=None):
         if res == 0:
             return None
         boxed_obj = self.api.jl_get_field(void_p(res), b'o')
--- a/julia/core.py
+++ b/julia/core.py
@@ -420,11 +420,12 @@ class Julia(object):
             # self._debug("No Exception")
             self.api.jl_exception_clear()
             return
-        self._debug("Retrieving exception infos...")
-        stderr = self.api.jl_stderr_obj()
-        self._debug("libjulia stderr = " + str(stderr))
-        self.api.jl_call2(self.api.show, stderr, exoc)
-        self._debug("show called ...")
+        if self.is_debugging:
+            self._debug("Retrieving exception infos...")
+            stderr = self.api.jl_stderr_obj()
+            self._debug("libjulia stderr = " + str(stderr))
+            self.api.jl_call2(self.api.show, stderr, exoc)
+            self._debug("show called ...")
         # self.api.jl_printf(self.api.jl_stderr_stream(), "\n");
 
         exception_type = self.api.jl_typeof_str(exoc).decode('utf-8')
--- a/julia/core.py
+++ b/julia/core.py
@@ -420,13 +420,6 @@ class Julia(object):
             # self._debug("No Exception")
             self.api.jl_exception_clear()
             return
-        if self.is_debugging:
-            self._debug("Retrieving exception infos...")
-            stderr = self.api.jl_stderr_obj()
-            self._debug("libjulia stderr = " + str(stderr))
-            self.api.jl_call2(self.api.show, stderr, exoc)
-            self._debug("show called ...")
-        # self.api.jl_printf(self.api.jl_stderr_stream(), "\n");
 
         # If, theoretically, an exception happens in early stage of
         # self.add_module_functions("Base"), showerror and sprint as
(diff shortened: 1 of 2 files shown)


stevengj:
I don't agree with this change.   If you evaluate Julia code and it throws an exception, then the exception should be shown — in fact, it should probably be converted to a Python exception and thrown in Python.

If import is printing an excessive number of exception messages, then import should be fixed.

tkf:
> it should probably be converted to a Python exception and thrown in Python

It *is* raised as Python exception at the end of `Julia.check_exception`, as I noted in #159:

https://github.com/JuliaPy/pyjulia/blob/2eb704c33d6f639b41300023e16f910b5945600e/julia/core.py#L426-L432

This is just below the code I edited.

The quirk in pyjulia I reported in #159 is that the error in Julia reported *twice*: first via `Base.show` and *then* Python exception.

tkf:
So I implemented what you suggested but actually I'm not sure if it is the best way to go, since, in principle, `check_exception` can now have infinite recursion (and so a stackoverflow).  This is because `self.sprint` and `self.showerror` may in turn call `check_exception`.  I know this is very unlikely but feels a bit unsafe.

I think a better approach would be to define a helper Julia function/macro that evals the code and wraps the result into a "Result" struct, something like

```jl
struct Result{T}
    result :: Nullable{T}
    error :: String
end
```

where `.result` is null and `.error` is set to `showerror` of the exception, when the code throws.

tkf:
(Actually, retrieving `Nullable` requires Julia function calls so maybe just let `.error` to be `nothing` or a `String`.  It doesn't have to type-stable when talking to Python.)

julia/core.py
```diff
@@ -428,9 +428,13 @@ def check_exception(self, src=None):
             self._debug("show called ...")
         # self.api.jl_printf(self.api.jl_stderr_stream(), "\n");
 
-        exception_type = self.api.jl_typeof_str(exoc).decode('utf-8')
+        res = self.api.jl_call2(void_p(self.api.convert), void_p(self.api.PyObject), void_p(exoc))
```
stevengj: What is this check for?

julia/core.py
```diff
@@ -428,9 +428,13 @@ def check_exception(self, src=None):
...
+        if res is None:
+            exception = self.api.jl_typeof_str(exoc).decode('utf-8')
+        else:
+            exception = self.sprint(self.showerror, self._as_pyobj(res))
```
stevengj: Why is `_as_pyobj` needed here?

stevengj: Oh, nevermind, I see that `exoc` is a raw Julia pointer.

julia/core.py
```diff
@@ -420,11 +420,12 @@ def check_exception(self, src=None):
...
-        self._debug("Retrieving exception infos...")
-        stderr = self.api.jl_stderr_obj()
-        self._debug("libjulia stderr = " + str(stderr))
-        self.api.jl_call2(self.api.show, stderr, exoc)
-        self._debug("show called ...")
+        if self.is_debugging:
+            self._debug("Retrieving exception infos...")
+            stderr = self.api.jl_stderr_obj()
+            self._debug("libjulia stderr = " + str(stderr))
+            self.api.jl_call2(self.api.show, stderr, exoc)
```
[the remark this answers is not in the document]
stevengj: I get the impression that the only point of this call to `self.api.show` in the first place was to make sure that the whole Julia exception is printed — maybe this code can simply be deleted now that `showerror` is included in the Python exception?

julia/core.py
```diff
@@ -420,16 +420,21 @@ def check_exception(self, src=None):
...
+            self._debug("show called ...")
         # self.api.jl_printf(self.api.jl_stderr_stream(), "\n");
 
-        exception_type = self.api.jl_typeof_str(exoc).decode('utf-8')
+        res = self.api.jl_call2(void_p(self.api.convert), void_p(self.api.PyObject), void_p(exoc))
```
stevengj: Probably should be `self.api.jl_call1(void_p(self.api.PyObject), void_p(exoc))` (the `convert` method may go away in a future PyCall version, since this is really a constructor call)

stevengj: I'm a little worried about the interaction of this raw `jl_value_t*` manipulation with [garbage collection in Julia](https://docs.julialang.org/en/release-0.6/manual/embedding/#Memory-Management-1), but I guess that's something that needs to be fixed more generally.

tkf: Reading git blame, it looks like it is changed from `self.api.jl_call1(void_p(self.api.PyObject), void_p(ans))` to `self.api.jl_call2(void_p(self.api.convert), void_p(self.api.PyObject), void_p(ans))` in https://github.com/JuliaPy/pyjulia/pull/54/commits/0912ac46f5e190b8ae6a64e675fa991b07daeb3c#diff-2f4f81856f2b92c6a4bb1e5bba2cf28aR380 by @Keno (PR #54). Not clear why from git message.

stevengj: Maybe `jl_call1` only works for actual `Function` objects these days, and we'll have to dig to find out how to call the constructor.   Let's continue to use `convert` for now, I guess.

tkf:
@stevengj Are there any other issues to be solved before merging it?

stevengj:
As mentioned above, I think the `"Retrieving exception infos"` debugging output code can be deleted — it is redundant now that the exception output is included in the Python exception.

tkf:
Sorry, I didn't notice that comment.  I forgot GitHub issue is not linearizing the comments...

I thought it could be useful in some edge cases.  For example, if an error occurred during `self.add_module_functions("Base")`, attributes `self.sprint` and `self.showerror` may not be set.

(...which makes me realize that I should make sure that they exist before using it; tweaked the code)

Having said that, I'm not the one maintaining the code so I'm happy with removing `if self.is_debugging:` block I added.  I did it in the last commit.

stevengj:
I guess CI is broken right now, but is it possible to add a test? For example, sqrt(-1) throws.

tkf:
There is already a test:

https://github.com/JuliaPy/pyjulia/blob/2eb704c33d6f639b41300023e16f910b5945600e/test/test_core.py#L27-L32

I tested it locally.

tkf:
I made the test a bit more specific to make sure we call `showerror`.

stevengj:
Thanks!

tkf:
Thanks!

Issue #132, opened 2017-11-08
IPython shell segfault on tab completion


simonbyrne:
In the ipython shell:
```
import julia
j = julia.Julia()
j.<tab>
```
immediately segfaults.

On OS X:
```
Python 3.6.3 (default, Oct  4 2017, 06:09:05) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.0.0 -- An enhanced Interactive Python. Type '?' for help.
```

simonbyrne:
Interestingly, it's fine in the normal python3 shell. So maybe it's an IPython issue?

stevengj:
IPython does tab completion differently than python3, but it's probably ultimately a pyjulia bug that is being triggered.

zsunberg:
On linux with ipython 5.1.0, I am able to do tab completion of methods of the Julia object, but when I try to complete filenames in a string, it segfaults, e.g.
```
In [1]: import julia

In [2]: j = julia.Julia()

In [3]: j.evalfile("new<tab>
signal (11): Segmentation fault                                               
while loading no file, in expression starting on line 0                       
Segmentation fault
```

tkf:
As @takluyver guessed in https://github.com/ipython/ipython/issues/10899#issuecomment-344541306, it looks like threading problem.  Just in case if someone wants to reproduce the problem in a script, here it is:

```python
import threading

from IPython.testing.globalipapp import get_ipython
from julia import Julia

ip = get_ipython()
ip.user_ns['julia'] = Julia()


def ip_complete():
    print(ip.complete('julia.sin.'))


print("ip_complete() in main thread:")
ip_complete()
print()

print("ip_complete() in different thread:")
th = threading.Thread(target=ip_complete)
th.start()
th.join()
```

It prints

```
ip_complete() in main thread:
('julia.sin.', ['julia.sin.jl_value'])

ip_complete() in different thread:

signal (11): Segmentation fault
```

I wonder if PyCall/libjulia can support multi-thread usage. https://github.com/JuliaLang/julia/issues/17573

tkf:
Of course, producing segfault this way doesn't require IPython:

```python
import threading

from julia import Julia

j = Julia()
j_sin = j.sin


def access_julia():
    j_sin.jl_value


th = threading.Thread(target=access_julia)
th.start()
th.join()
```

takluyver:
[...]

PR #162, opened 2018-05-15
Make module loading lazy

tkf:
@stevengj suggested to make JuliaModule lazy: https://github.com/JuliaPy/pyjulia/issues/159#issuecomment-385193887

Here is my implementation for that.  I actually added some more changes on top that to solve relevant issues.  If some components of the changes have to be removed, let me know.  Those components are implemented roughly in the following order:

### <strike>Basing on CI fix #149</strike>

<strike>
First of all this PR is based on #149 to run CI. Let me know if I need to rebase once #149 is merged.
</strike>

(Edited: now rebased)

### Lazy module member loading

`Julia.eval` is called via `JuliaModule.__getattr__` so that Julia object is brought to Python world only when it is requested.

Implemented in https://github.com/JuliaPy/pyjulia/pull/162/commits/dc40166351fd32d91668f0eacdda4a6535fc0e37 and finished by https://github.com/JuliaPy/pyjulia/pull/162/commits/843b51959a50d8cff9e07b48b1c0032408dad532

### Julia's interpreter global namespace as a module `julia.Main`

In addition to usual `JuliaModule`, the module class `JuliaMainModule` of `julia.Main` has an implementation of `__setattr__` so that you can send Python object to Julia by just setting it:

```python
>>> from julia import Main
>>> Main.xs = [1, 2, 3]
```

Since it supersede `core.Julia`, I suggest to deprecate accessing Julia values via `core.Julia().<name>`: https://github.com/JuliaPy/pyjulia/pull/162/commits/b30af4541296bfcce2f838352cd2a6f0cfe90560

### Automatically initialize Julia with `from julia import ...`

This makes `from julia import <Julia module>` works without the initial setup.  Note that a custom setup can still be done by calling `julia.Julia` with appropriate arguments *before* trying to import Julia modules:
https://github.com/JuliaPy/pyjulia/pull/162/commits/51997c0e4449d9dad330d34386d2caca1dfdbe7f

See also the document on the new API:
https://github.com/tkf/pyjulia/blob/lazy-module/README.md#usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,23 +58,70 @@ If you run into problems using `pyjulia`, first check the version of `PyCall.jl`
 
 Usage
 -----
-To call Julia functions from python, first import the library
+
+`pyjulia` provides a high-level interface which assumes a "normal"
+setup (e.g., `julia` is in your `PATH`) and a low-level interface
+which can be used in a customized setup.
+
+### High-level interface
+
+To call a Julia function in Julia module, import the Julia module (say
+`Base`) by:
 
 ```python
-import julia
+from julia import Base
 ```
 
-then create a Julia object that makes a bridge to the Julia interpreter (assuming that `julia` is in your `PATH`)
+and then call Julia functions in `Base` from python, e.g.,
 
 ```python
-j = julia.Julia()
+Base.sind(90)
 ```
 
-You can then call Julia functions from python, e.g.
+Other variants of Python import syntax also work:
 
 ```python
-j.sind(90)
+import julia.Base
+from julia.Base import LinAlg   # import a submodule
+from julia.Base import sin      # import a function from a module
 ```
+
+The global namespace of Julia's interpreter can be accessed via a
+special module `julia.Main`:
+
+```python
+from julia import Main
+```
+
+You can set names in this module to send Python values to Julia:
+
+```python
+Main.xs = [1, 2, 3]
+```
+
+so that, e.g., it can be evaluated at Julia side using Julia syntax:
+
+```python
+Main.eval("sin.(xs)")
+```
+
+### Low-level interface
+
+If you need a custom setup for `pyjulia`, it must be done *before* any
+imports of Julia modules.  For example, to use Julia interpreter at
+`PATH/TO/MY/CUSTOM/julia`, run:
+
+```python
+from julia import Julia
+j = julia.Julia(jl_runtime_path="PATH/TO/MY/CUSTOM/julia")
+```
+
+You can then use, e.g.,
+
+```python
+from julia import Base
+```
+
 
 How it works
 ------------
--- a/README.md
+++ b/README.md
@@ -65,8 +65,8 @@ which can be used in a customized setup.
 
 ### High-level interface
 
-To call a Julia function in Julia module, import the Julia module (say
-`Base`) by:
+To call a Julia function in a Julia module, import the Julia module
+(say `Base`) with:
 
 ```python
 from julia import Base
@@ -99,7 +99,8 @@ You can set names in this module to send Python values to Julia:
 Main.xs = [1, 2, 3]
 ```
 
-so that, e.g., it can be evaluated at Julia side using Julia syntax:
+which allows it to be accessed directly from Julia code, e.g., it can
+be evaluated at Julia side using Julia syntax:
 
 ```python
 Main.eval("sin.(xs)")
@@ -107,9 +108,9 @@ Main.eval("sin.(xs)")
 
 ### Low-level interface
 
-If you need a custom setup for `pyjulia`, it must be done *before* any
-imports of Julia modules.  For example, to use Julia interpreter at
-`PATH/TO/MY/CUSTOM/julia`, run:
+If you need a custom setup for `pyjulia`, it must be done *before*
+importing any Julia modules.  For example, to use the Julia
+interpreter at `PATH/TO/MY/CUSTOM/julia`, run:
 
 ```python
 from julia import Julia
--- a/julia/core.py
+++ b/julia/core.py
@@ -56,6 +56,12 @@ def jl_name(name):
     return name
 
 
+def py_name(name):
+    if name.endswith('!'):
+        return name[:-1] + '_b'
+    return name
+
+
 class JuliaModule(ModuleType):
 
     def __init__(self, loader, *args, **kwargs):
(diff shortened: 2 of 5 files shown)


julia/core.py
```diff
@@ -46,7 +46,45 @@ class JuliaError(Exception):
...
+        try:
+            module_path = ".".join((juliapath, name))
+            is_module = self.__julia.eval("isa({}, Module)"
+                                          .format(module_path))
+            if is_module:
+                # TODO: find a better way to handle submodules
+                split_path = module_path.split(".")
+                is_base = split_path[-1] == "Base"
+                recur_module = split_path[-1] == split_path[-2]
+                if not is_base and not recur_module:
```
tkf: Not sure what is the best way to handle submodules.  Is there an easy way to check if some module is a submodule of other module in Julia?  If not, it's probably better to let user explicitly import submodules (i.e., `julia.Module1.Module2` is an `AttributeError` until the user does `import julia.Module1.Module2`)?

tkf:
In the current master, we have `"jl".join(name)` in `JuliaModuleLoader` when Julia name is a Python keyword:

https://github.com/JuliaPy/pyjulia/blob/7ebad2276c932083cdca456951be8e183ff9969f/julia/core.py#L77-L88

What was the intention of this code?  I mean, `"jl".join("with")` evaluates to `'wjlijltjlh'` which is not a very human-friendly name.

Was it probably not by intention?  Was it maybe intended to be `"jl" + name` or something?

At the moment, this PR has no code for Python keyword handling.  We can add something like:

```python
if name.startswith("jl_"):
    try:
        return self.__try_getattr(name[len("jl_"):])
    except AttributeError:
        pass
```

test/test_core.py
```diff
@@ -61,9 +66,61 @@ def test_import_julia_functions(self):
...
+        assert 'resize_b' in Base.__all__
+
+    def test_module_dir(self):
+        from julia import Base
+        assert 'resize_b' in dir(Base)
+
+    def test_import_without_setup(self):
+        command = [sys.executable, '-c', 'from julia import Base']
+        print('Executing:', *command)
+        subprocess.check_call(command, env=_orig_env)
```
tkf: `env=_orig_env` probably is required to fix https://travis-ci.org/JuliaPy/pyjulia/jobs/379514563#L298

tkf: Yes it looks like it fixes the failure: https://travis-ci.org/JuliaPy/pyjulia/builds/379524930

rdeits:
CI is now fixed, so you might want to rebase this PR

tkf:
Thanks. Rebased.

julia/core.py
```diff
@@ -40,70 +41,132 @@ def iteritems(d): return iter(d.items())
...
+                try:
+                    return self.__try_getattr(jl_name(name))
+                except AttributeError:
+                    pass
+            raise
+
+    def __try_getattr(self, name):
+        juliapath = remove_prefix(self.__name__, "julia.")
+        try:
+            module_path = ".".join((juliapath, name))
```
stevengj: maybe `qualified_name` or something?   Calling this a "path" is a bit confusing.

julia/core.py
```diff
@@ -40,70 +41,132 @@ def iteritems(d): return iter(d.items())
...
+                # TODO: find a better way to handle submodules
+                split_path = module_path.split(".")
+                is_base = split_path[-1] == "Base"
+                recur_module = split_path[-1] == split_path[-2]
+                if not is_base and not recur_module:
+                    newpath = ".".join((self.__name__, name))
+                    return self.__loader__.load_module(newpath)
+            return self._julia.eval(module_path)
+        except JuliaError:
+            if isafunction(self._julia, name, mod_name=juliapath):
```
stevengj: I'm confused by this `isafunction` check.   Didn't it already try `self._julia.eval(module_path)` for all objects that are not modules, including functions?   And if that threw an error, shouldn't we rethrow it?

julia/core.py
```diff
@@ -40,70 +41,132 @@ def iteritems(d): return iter(d.items())
...
+            if name.endswith("_b"):
+                try:
+                    return self.__try_getattr(jl_name(name))
+                except AttributeError:
+                    pass
+            raise
+
+    def __try_getattr(self, name):
+        juliapath = remove_prefix(self.__name__, "julia.")
+        try:
```
stevengj: instead of this `try` block, maybe we should try calling `isdefined(module, :name)`, throwing an `AttributeError` if it returns `false`, and letting any `JuliaError` that is thrown fall through to the caller.

julia/core.py
```diff
@@ -140,8 +140,6 @@ class JuliaImporter(object):
 
     # find_module was deprecated in v3.4
     def find_module(self, fullname, path=None):
-        if path is None:
```
tkf: It looks like this code was added in https://github.com/JuliaPy/pyjulia/commit/fa157b19e6fcd6fdcdd2f0e7584e6d9c75cc091b and carried around as-is.

Probably just a remnant of experimental code?

tkf:
Hi @kmsquire, thanks a lot for correcting my English!

BTW, it's interesting that you find that IPython tab completion works in pyjulia master.  I thought, in theory, pyjulia would never work with IPython < 7.0 since the tab-completion is done in different thread https://github.com/JuliaPy/pyjulia/issues/132#issuecomment-390474080 and pyjulia (or rather libjulia, I guess) is not thread-safe.  I actually find out that IPython master (= 7.x) already works in single-thread just now.  I have to check it...

README.md
```diff
@@ -58,24 +58,71 @@ If you run into problems using `pyjulia`, first check the version of `PyCall.jl`
...
+from julia import Main
 ```
 
+You can set names in this module to send Python values to Julia:
+
+```python
+Main.xs = [1, 2, 3]
+```
+
+so that, e.g., it can be evaluated at Julia side using Julia syntax:
```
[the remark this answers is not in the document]
tkf: I totally agree that communication-by-sharing is a terrible idea and using this feature in Python modules should be forbidden :)  But I was thinking in terms of interactive usage, mostly via `%julia`.  In that case, passing around code via global namespace is pretty useful.

tkf:
Test failure is due to CI setting. See #171.

stevengj:
Is this ready to merge?  I haven't really been keeping up with pyjulia…

tkf:
I *think* so.  I mean, I used this branch for a few weeks just after creating it and I don't remember encountering any regressions.  After that, the only change was in README.  I'm not planning to add anything more to this PR.

tkf:
Rebased onto master. Travis and AppVeyor are all green now.

PR #167, opened 2018-06-20
Improve julia v0.7 support

rdeits:
Builds on #166 to fix a few more v0.7 issues


rdeits:
Travis and appveyor are both green. Merging!

PR #169, opened 2018-06-24
Updating installation guide

jlperla:
Makes the importance of `ENV["PYTHON"]` clear, and points out that you need to relate the python interpreter and the julia executale

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,17 +16,30 @@ python -m unittest discover
 
 `pyjulia` is tested against Python versions 2.7 and 3.5.  Older versions of Python (than 2.7)  are not supported.
 
-Installation
-------------
-You will need to install PyCall in your existing Julia installation
+Installation of PyCall
+------------------------
+PyCall needs to be associated with the python executable with which you will call pyjulia.  To find the executable location from within a python interpret
+```python
+>>> import os
+>>> import sys
+>>> os.path.dirname(sys.executable)
+'C:\\Anaconda3'
+```
+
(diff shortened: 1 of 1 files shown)


stevengj:
I thought there was support for using a different PyCall configuration in Julia vs pyjulia; am I misremembering?

jlperla:
Might be... All I know is this was the source of all of my problems with the library

stevengj:
Supporting a pyjulia-specific PyCall configuration is a key feature — otherwise, it will be impossible to install pyjulia into more than one Python environment on the same machine.

stevengj:
See also https://github.com/JuliaPy/pyjulia/tree/master/julia/fake-julia

stevengj:
@Keno, has something bitrotted here?

tkf:
> support for using a different PyCall configuration in Julia vs pyjulia

I just noticed Travis CI with `CROSS_VERSION=1` is put in `allow_failures` (since #149?): https://travis-ci.org/JuliaPy/pyjulia/builds/394733702 I suppose it's not working for a while.

But it looks like it's working on Windows: https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.90

A bit tangential but I wonder if Pkg3 can help pyjulia/PyCall.  There is a discussion on supporting 32 and 64 bit binaries by separating precompilation paths. https://github.com/JuliaLang/Pkg.jl/issues/93  Would it be possible to get rid of `fake-julia` using the same mechanism?  I'm thinking something like Pkg3-tracked environment per Python's virtualenv.

tkf:
This should already be fixed in master. Please open a new issue/PR if there is something to be improved.

PR #170, opened 2018-07-09
Test against latest 3.7

PallHaraldsson:


tkf:
It looks like 3.7 is not available in Travis yet? https://github.com/travis-ci/travis-ci/issues/9815

3.7-dev is already there though. https://docs.travis-ci.com/user/languages/python/

tkf:
FYI I'm doing some overhaul of the test suite in #178 which makes this kind of change easier (for Travis CI; for AppVeyor, not so match...).

tkf:
We are testing with Python 3.7 in macOS and I guess it's OK for now.  Our build matrix is rather large already.

PR #178, opened 2018-08-01
WIP Use tox and pytest

tkf:
It simplifies local development and CI setup:

* Cross-version testing can be done with a single command locally.
* We can make sure that same set of Python packages are installed locally and in CI clouds.
* `tox` and `pytest` outputs are much nicer and helpful (but there are options for verbose outputs)

It's WIP since I'm not sure Travis and AppVeyor setups are correct yet.


tkf:
So it looks like it is not testing "cache reuse" mode at the moment.  Let's see if I can find a nice way to fix this...

tkf:
OK.  I think it's ready for review. (ping @stevengj @rdeits)

To support "shared-cache mode", I added a script `julia/with_rebuilt.py` to build PyCall.jl with the Python interpreter used for testing and build back to the original configuration.  This is documented in README too: https://github.com/tkf/pyjulia/tree/tox#testing

Some summaries for misc details:

* I switched to testing against nightly to 0.7-rc to wait for `PyCall.jl` etc to get ready for 1.0.  This required a slight improvement of `ci/install-julia.sh`.
* I added tests for IPython magics and fixed some bugs in the module importer.  I pulled #151 and added tests for it too.
* In AppVeyor, tests for Python 2 and 3 are run in the same job.  Considering low-throughput of AppVeyor and overhead of downloading and installing Julia, I think it's a reasonable compromise.  But we can always separate them in different jobs by adding `TOXENV` for each job.

### Some details for cross-version test in Windows with Python

Note that cross-version test in Windows with Python 2 fails and the failure is ignored at the moment.  But it turned out our previous CI setting didn't actually test cross-version capability.  The problematic part was here:

https://github.com/JuliaPy/pyjulia/blob/9dfc132474dcf2d0f7ab4fa1ff06ca8176ec94c7/appveyor.yml#L79

There were two issues: (1) `CROSS_VERSION_PATH` was not actually path to the executable.  It perhaps meant to be `$env:CROSS_VERSION_PATH\\python -m unittest discover` or something.  (2) `Invoke-Expression` does not fail even when the invoked command fails.  So the tests were passing even though we had an error: https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.107/job/8rhimj9mkxins0yn#L101

I think the way Python 2 cross-version test fails needs attention from PyCall.jl/pyjulia experts.  The error happens when evaluating `using PyCall` and it says:

```
LLVM ERROR: Program used external function 'trunc' which could not be resolved!
```

(see, e.g., https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.134/job/0l9qvw8dgp1636s4#L158)

I remember having the same issue when I helped setting up CI for diffeqpy https://github.com/JuliaDiffEq/diffeqpy/issues/25  It used the released version of pyjulia so it's possible that this problem has been there for some times.

julia/core.py
```diff
@@ -141,6 +142,10 @@ class JuliaImporter(object):
     # find_module was deprecated in v3.4
     def find_module(self, fullname, path=None):
         if fullname.startswith("julia."):
+            pypath = os.path.join(os.path.dirname(__file__),
+                                  "{}.py".format(fullname[len("julia."):]))
+            if os.path.isfile(pypath):
```
tkf: We need to fallback to standard Python importer in this case to import Python submodules (e.g., `julia/magic.py`).

tkf: Of course, this has to be improved if/when we have python subpackages.

julia/magic.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, shell):
               end='')
         # Flush, otherwise the Julia startup will keep stdout buffered
         sys.stdout.flush()
-        self.julia = Julia(init_julia=True)
+        self._julia = Julia(init_julia=True)
```
tkf: Previous `self.julia = ...` was shadowing the method `julia`.  It was fine because the method was actually not called directly in IPython magic system.  But it's handy to not shadow for testing.

ci/appveyor/win64/python3.5.bat
```diff
@@ -0,0 +1 @@
+@C:\Python35-x64\python.exe %*
```
tkf: I made ` ci/appveyor/win*/python*.bat` files, following the recommendation here:

https://tox.readthedocs.io/en/latest/developers.html#id2
https://github.com/tox-dev/tox/issues/114#issuecomment-349712050

Note that adding CI support for Python 3.6 and 3.7 requires adding more of those files.

tkf:
@stevengj Can I merge my PR without a review? Some OSS projects have the policy that you don't merge your own PR. Not sure how it works in JuliaPy. I wonder something like "if my PR does not induce major or minor version bumps, it's likely to be OK to pull it myself" would work. I wouldn't still pull it myself if it's a huge change in the entire codebase like a drastic refactoring.

PR #183, opened 2018-08-17
Julia 1.0 support

tkf:
It is based on #178.  It's probably better to pull #178 first to make reviewing easier.

### DIFF
--- a/julia/core.py
+++ b/julia/core.py
@@ -342,7 +342,7 @@ class Julia(object):
                 runtime = jl_runtime_path
             else:
                 runtime = 'julia'
-            JULIA_HOME, libjulia_path, image_file, depsjlexe = juliainfo()
+            JULIA_HOME, libjulia_path, image_file, depsjlexe = juliainfo(runtime)
             self._debug("pyprogramname =", depsjlexe)
             self._debug("sys.executable =", sys.executable)
             exe_differs = is_different_exe(depsjlexe, sys.executable)
--- a/julia/core.py
+++ b/julia/core.py
@@ -262,14 +262,16 @@ def juliainfo(runtime='julia'):
         [runtime, "-e",
          """
          println(VERSION < v"0.7.0-DEV.3073" ? JULIA_HOME : Base.Sys.BINDIR)
+         if VERSION >= v"0.7.0-DEV.3630"
+             using Libdl
+             using Pkg
+         end
          println(Libdl.dlpath(string("lib", splitext(Base.julia_exename())[1])))
          println(unsafe_string(Base.JLOptions().image_file))
          PyCall_depsfile = Pkg.dir("PyCall","deps","deps.jl")
          if isfile(PyCall_depsfile)
-            eval(Module(:__anon__),
-                Expr(:toplevel,
-                 :(Main.Base.include($PyCall_depsfile)),
-                 :(println(pyprogramname))))
+             include(PyCall_depsfile)
+             println(pyprogramname)
          end
          """])
     args = output.decode("utf-8").rstrip().split("\n")
--- a/julia/core.py
+++ b/julia/core.py
@@ -274,7 +274,7 @@ def juliainfo(runtime='julia'):
              modpath = Base.locate_package(Base.identify_package("PyCall"))
              PyCall_depsfile = joinpath(dirname(modpath),"..","deps","deps.jl")
          end
-         if isfile(PyCall_depsfile)
+         if PyCall_depsfile !== nothing && isfile(PyCall_depsfile)
              include(PyCall_depsfile)
              println(pyprogramname)
          end
--- a/julia/core.py
+++ b/julia/core.py
@@ -268,7 +268,12 @@ def juliainfo(runtime='julia'):
          end
          println(Libdl.dlpath(string("lib", splitext(Base.julia_exename())[1])))
          println(unsafe_string(Base.JLOptions().image_file))
-         PyCall_depsfile = Pkg.dir("PyCall","deps","deps.jl")
+         if VERSION < v"0.7.0"
+             PyCall_depsfile = Pkg.dir("PyCall","deps","deps.jl")
+         else
+             modpath = Base.locate_package(Base.identify_package("PyCall"))
+             PyCall_depsfile = joinpath(dirname(modpath),"..","deps","deps.jl")
+         end
          if isfile(PyCall_depsfile)
              include(PyCall_depsfile)
              println(pyprogramname)
--- a/julia/with_rebuilt.py
+++ b/julia/with_rebuilt.py
@@ -51,6 +51,18 @@ def maybe_rebuild(rebuild, julia):
 
 @contextmanager
 def ignoring(sig):
+    """
+    Context manager for ignoring signal `sig`.
+
+    For example,::
+
+        with ignoring(signal.SIGINT):
+            do_something()
+
+    would ignore user's ctrl-c during ``do_something()``.  This is
+    useful when launching interactive program (in which ctrl-c is a
+    valid keybinding) from Python.
+    """
     s = signal.signal(sig, signal.SIG_IGN)
     try:
         yield
--- a/julia/with_rebuilt.py
+++ b/julia/with_rebuilt.py
@@ -8,6 +8,7 @@ variable `PYJULIA_TEST_REBUILD` is set to ``yes``.
 from __future__ import print_function, absolute_import
 
 import os
+import signal
 import subprocess
 import sys
 from contextlib import contextmanager
@@ -42,8 +43,17 @@ def maybe_rebuild(rebuild, julia):
         yield
 
 
+@contextmanager
+def ignoring(sig):
+    s = signal.signal(sig, signal.SIG_IGN)
+    try:
+        yield
+    finally:
+        signal.signal(sig, s)
+
+
 def with_rebuilt(rebuild, julia, command):
-    with maybe_rebuild(rebuild, julia):
+    with maybe_rebuild(rebuild, julia), ignoring(signal.SIGINT):
         print('Execute:', *command)
         return subprocess.call(command)
 
--- a/julia/with_rebuilt.py
+++ b/julia/with_rebuilt.py
@@ -22,7 +22,12 @@ def maybe_rebuild(rebuild, julia):
         env = os.environ.copy()
         info = juliainfo(julia)
 
-        build = [julia, '-e', 'Pkg.build("PyCall")']
+        build = [julia, '-e', """
+        if VERSION >= v"0.7.0-DEV.3630"
+            using Pkg
+        end
+        Pkg.build("PyCall")
+        """]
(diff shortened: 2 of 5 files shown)


tkf:
Awesome!  Thanks for the merge!  It's rebased now.

tkf:
CIs are all green now!

mprogram:
Why then do we have:
```
test/test_core.py exception occured? None
exception occured? None
.exception occured? 139909767752752
...exception occured? None
exception occured? None
.exception occured? None
exception occured? None
exception occured? None
exception occured? None
exception occured? None
.......Executing: /home/travis/build/JuliaPy/pyjulia/.tox/py/bin/python -c from julia import Base
┌ Warning: `Pkg.dir(pkgname, paths...)` is deprecated; instead, do `import PyCall; joinpath(dirname(pathof(PyCall)), "..", paths...)`.
└ @ Pkg.API /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:454
...exception occured? None
.....                                         
test/test_magic.py Initializing Julia interpreter. This may take some time...
.Initializing Julia interpreter. This may take some time...
.Initializing Julia interpreter. This may take some time...
.Initializing Julia interpreter. This may take some time...
Exception 'ArgumentError' occurred while calling julia code:
pop!([])
.Initializing Julia interpreter. This may take some time...
Exception 'ErrorException' occurred while calling julia code:
1 += 1
.                                                       
```
in the Travis log (checked that of Python 3.6)?

julia/core.py
```diff
@@ -262,14 +262,16 @@ def juliainfo(runtime='julia'):
...
          println(Libdl.dlpath(string("lib", splitext(Base.julia_exename())[1])))
          println(unsafe_string(Base.JLOptions().image_file))
          PyCall_depsfile = Pkg.dir("PyCall","deps","deps.jl")
          if isfile(PyCall_depsfile)
-            eval(Module(:__anon__),
-                Expr(:toplevel,
-                 :(Main.Base.include($PyCall_depsfile)),
-                 :(println(pyprogramname))))
+             include(PyCall_depsfile)
+             println(pyprogramname)
```
stevengj: I thought the reason for putting this in an anonymous module was to avoid putting `pyprogramname` into the global namespace.  Is this no longer a concern?

julia/with_rebuilt.py
```diff
@@ -42,8 +49,17 @@ def maybe_rebuild(rebuild, julia):
         yield
 
 
+@contextmanager
+def ignoring(sig):
+    s = signal.signal(sig, signal.SIG_IGN)
+    try:
+        yield
+    finally:
+        signal.signal(sig, s)
```
stevengj: Can we have a comment about what's going on here?

test/test_core.py
```diff
@@ -97,8 +97,8 @@ def test_from_import_non_existing_julia_name(self):
     def test_julia_module_bang(self):
         from julia import Base
         xs = [1, 2, 3]
-        ys = Base.scale_b(xs[:], 2)
-        assert all(x * 2 == y for x, y in zip(xs, ys))
+        ys = Base.fill_b(xs[:], 10)
```
stevengj: Why not `ys = Base.fill_b(xs, 10)` ?

julia/core.py
tkf: I thought it was Julia 0.4 compatibility or something. This is executed in a subprocess which exits right after printing `pyprogramname`. So I don't think we have namespace problem. It is added in https://github.com/JuliaPy/pyjulia/commit/fde40d702a42a4aa5dda5dac873cbefbcc97baf3#diff-2f4f81856f2b92c6a4bb1e5bba2cf28aR260

tkf:
@mprogram Those are expected debug messages. We pass `-s` option to `pytest` so that it's easier to debug CI but I agree it's not very clean.

test/test_core.py
stevengj: I realize that, but the new test doesn't actually test much.  That's why I suggested just setting `ys = [2, 4, 6]` and keeping the test of `all` as-is.

tkf: My intention when writing the test was just to test variable renaming code path (back then I wasn't sure if it was OK to import numpy).  But now that we have it via tox, I suggest something like:

```python
xs = numpy.zeros(3)
Base.fill_b(xs, 10)
assert all(ys == 10)
```

or

```python
xs = numpy.array([1, 2, 3])
ys = xs[:]
if julia.eval("VERSION >= v"0.7-""):
    Base.rmul_b(xs, 2)
else:
    Base.scale_b(xs, 2)
assert all(2 * ys == xs)
```

Which one do you like?

julia/with_rebuilt.py
tkf: I added a docstring.

noahstier:
With `julia --version` -> `1.0.0` and `python --version` -> `Python 3.5.5 :: Anaconda, Inc.`, if I do this:
```
pip install git+https://github.com/tkf/pyjulia.git
python -c "import julia; julia.Julia()"
```
I get this error:
```
>>> julia.Julia()
┌ Warning: `Pkg.dir(pkgname, paths...)` is deprecated; instead, do `import PyCall; joinpath(dirname(pathof(PyCall)), "..", paths...)`.
└ @ Pkg.API /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:454
  File "~/miniconda3/envs/35/lib/python3.5/site-packages/julia/core.py", line 562, in __init__
    self.__julia = Julia(*args, **kwargs)
  File "~/miniconda3/envs/35/lib/python3.5/site-packages/julia/core.py", line 440, in __init__
    "\"lib\", \"pyjulia%s-v$(VERSION.major).$(VERSION.minor)\"))" % sys.version_info[0])
  File "~/miniconda3/envs/35/lib/python3.5/site-packages/julia/core.py", line 488, in _call
    self.check_exception(src)
  File "~/miniconda3/envs/35/lib/python3.5/site-packages/julia/core.py", line 517, in check_exception
    .format(exception, src))
julia.core.JuliaError: Exception 'UndefVarError' occurred while calling julia code:
pushfirst!(Base.LOAD_CACHE_PATH, abspath(Pkg.Dir._pkgroot(),"lib", "pyjulia3-v$(VERSION.major).$(VERSION.minor)"))
```

Investigating:
`julia -e "println(Base.LOAD_CACHE_PATH)"`
gives
`ERROR: UndefVarError: LOAD_CACHE_PATH not defined`

and
`julia -e "using Pkg; println(Pkg.Dir)"`
gives
`ERROR: UndefVarError: Dir not defined`

tkf:
@noahstier #185

test/test_core.py
tkf: Actually, the test I proposed does not work. Strange...

```python
    def test_julia_module_bang(self):
        from julia import Base
        xs = numpy.zeros(3)
        ys = Base.fill_b(xs, 10)
        assert all(ys == 10)
>       assert all(xs == 10)
E       AssertionError: assert False
E        +  where False = all(array([0., 0., 0.]) == 10)
```

tkf: This happens in PyCall as well.  Is it a known issue or this is intended?

```julia
julia> py"""
       import numpy
       xs = numpy.zeros(3)
       """

julia> fill!(py"xs", 10)
3-element Array{Float64,1}:
 10.0
 10.0
 10.0

julia> py"xs"
3-element Array{Float64,1}:
 0.0
 0.0
 0.0

(v1.0) pkg> st
    Status `~/.julia/environments/v1.0/Project.toml`
  ...
  [438e738f] PyCall v1.18.0
  ...
```

mprogram:
@noahstier look at JuliaPy/pyjulia/issues/148 for a patching hint (presently blocked by #185)

test/test_core.py
tkf: Ah, that was because `py"xs"` was converted to `PyAny`.  If I do `fill!(PyArray(PyCall.maindict()["xs"]), 10)` in Julia, mutation works (= it's reflected on Python side).

I'm not sure what's the right illustration of this problem in Julia.  Something like this?:

```julia
julia> py"""
       def apply(f):
           xs = numpy.zeros(3)
           f(xs)
           return xs
       """

julia> py"apply"(xs -> fill!(xs, 10))
3-element Array{Float64,1}:
 0.0
 0.0
 0.0
```

stevengj: See https://github.com/JuliaPy/PyCall.jl/pull/531#issuecomment-414314622 … you need to specify the type conversion you want for an argument if you want in-place operations to be possible in general.

tkf: Got it. Yes, I think explicitly denoting which argument is mutating is a good idea.

So, for this test, we just need to check the returned value, right?:

```python
    def test_julia_module_bang(self):
        from julia import Base
        xs = [1, 2, 3]
        ys = Base.fill_b(xs, 10)
        assert all(y == 10 for y in ys)
```

Although it may be better if we just drop `_b` since users have to implement a callback on Julia side to avoid allocations.

tkf: OK. So I think I finally found a meaningful test:

```python
    def test_julia_module_bang(self):
        from julia.Base import Channel, put_b, take_b
        chan = Channel(1)
        sent = 123
        put_b(chan, sent)
        received = take_b(chan)
        assert sent == received
```

What do you think?

In the future, I think we need an API to denote some Python values have to be passed without copying.  Something like `fill_b(asref(arr), 10)`.

stevengj:
Ready to merge?

tkf:
What about https://github.com/JuliaPy/pyjulia/pull/183#discussion_r211484573? The problem you pointed about `test_julia_module_bang` is not solved yet. I can apply my last suggestion if you think that's better.

test/test_core.py
```diff
@@ -95,10 +97,12 @@ def test_from_import_non_existing_julia_name(self):
...
-        from julia import Base
-        xs = [1, 2, 3]
-        ys = Base.scale_b(xs[:], 2)
-        assert all(x * 2 == y for x, y in zip(xs, ys))
+        from julia.Base import Channel, put_b, take_b
+        chan = Channel(1)
+        sent = 123
+        put_b(chan, sent)
+        received = take_b(chan)
+        assert sent == received
```
tkf: I go ahead and I do https://github.com/JuliaPy/pyjulia/pull/183#discussion_r211484573 (to make sure that it works on CIs). If you prefer `fill_b` or something else let me know.

If CIs and you are happy with it then I think it's ready for merge.

tkf:
I wonder why appveyor is not running.  There is "Intermittent issues with Chocolatey packages installation" https://appveyor.statuspage.io/incidents/kqwjgg14m71g but I don't think it's related to our setup (unless it is used internally somewhere).

test/test_core.py
stevengj: I think the right thing to do is to make it easy to use the `pyfunction` method from Julia, for cases where you need control over precisely how arguments are passed.

PR #187, opened 2018-08-19
Resolve runtime path

tkf:
#183 needs to go first. It's still a WIP as I need to update the docs. But I want to check that it works on other platforms first.


stevengj:
Good to go once docs are updated?

tkf:
Rebased and added the docs. It's good to go now.

PR #188, opened 2018-08-19
Small cleanups

tkf:
Some small cleanups/fixes I noticed while writing #183 etc.  I'll rebase it after #183.


tkf:
Rebased. CIs are all green. @stevengj Please take a look at it.

PR #109, opened 2017-07-24
Update README.md

madeleineudell:
More instructions for troubleshooting.

Issue #182, opened 2018-08-12
How about checking Python version rather than the executable path?


tkf:
Currently, pyjulia tries to use separate cache if (1) the path of Python interpreters are different _or_ (2) if the current interpreter is statically linked:

https://github.com/JuliaPy/pyjulia/blob/9dfc132474dcf2d0f7ab4fa1ff06ca8176ec94c7/julia/core.py#L335

But according to @stevengj, all that matter is the major part of the Python version number https://github.com/JuliaPy/pyjulia/pull/172#issuecomment-409357199.  Then isn't it better to just compare the major version number (which is stored in `PyCall/deps/deps.jl`)?  This would alleviate the issue #173 (difficulty of private cache support in Julia >= 0.7) since this makes Python 3 distributed by some Linux distros (but sadly not Ubuntu) usable with the same PyCall cache.  For example, this would make pyjulia usable with Julia >= 0.7 in [python Docker image](https://hub.docker.com/_/python/).

If this is the right way to go, I can do the fix (which would be very straightforward, I think).

stevengj:
Actually, I think I was wrong — [this code](https://github.com/JuliaPy/PyCall.jl/blob/3e5831e45421cdedf16397e3653c308f7b4d68c0/src/startup.jl#L113-L139) will precompile differently if libpython changes (or if libpython is statically linked).

tkf:
Actually, what I meant to write was to replace the check to see if the path is different (`exe_differs`) with check for the major version number.  I wasn't suggesting to remove `determine_if_statically_linked()`.  (Re-reading it, this is totally not obvious --- sorry about that.)

I also just noticed that checking just major version is not enough since this code also would compile differently in Python < 3.2 and Python >= 3.2.  Am I correct?:

https://github.com/JuliaPy/PyCall.jl/blob/3e5831e45421cdedf16397e3653c308f7b4d68c0/src/startup.jl#L108

(Though I think it's OK to drop Python < 3.2.  Aren't they somewhat more exotic than 2.7 nowadays?)

So, does it make sense to replace `exe_differs` with major version check (or major + minor, if we support < 3.2)?

tkf:
Ah, I see. Thanks a lot for the explanation. So then my plan won't work...

Issue #175, opened 2018-08-01
A script for reproducing the initialization error


tkf:
I think I found a reproducible script to get the error I saw sometimes.  Confusingly, the key to **get** the error was to **execute** `using PyCall` in Julia to invoke precompilation.  Before running this script, remove `~/.juliarc.jl` if you have PyCall-related configuration.:

```bash
rm -rvf ~/.julia/lib/pyjulia{2,3}-v0.6 ~/.julia/lib/v0.6/PyCall.ji ~/.julia/v0.6/PyCall/deps/PYTHON ~/.julia/v0.6/PyCall/deps/deps.jl

PYTHON="$(which python3)" julia -e 'Pkg.build("PyCall"); using PyCall'
# PYTHON="$(which python3)" julia -e 'Pkg.build("PyCall")'  # this produces no error below

conda create --yes --prefix py37 python=3.7

source ~/*conda*/bin/activate ./py37
which python
python --version
python -c 'import julia; julia.Julia()'
```

This script produces:

```pytb
Error processing line 1 of /usr/lib/python3.6/site-packages/virtualenvwrapper-4.8.2-py3.6-nspkg.pth:

Failed to import the site module
Traceback (most recent call last):
  File "/usr/lib/python3.6/site.py", line 168, in addpackage
    exec(line)
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.6/types.py", line 171, in <module>
    import functools as _functools
  File "/usr/lib/python3.6/functools.py", line 21, in <module>
    from collections import namedtuple
  File "/usr/lib/python3.6/collections/__init__.py", line 29, in <module>
    import heapq as _heapq
  File "/usr/lib/python3.6/heapq.py", line 587, in <module>
    from _heapq import *
SystemError: initialization of _heapq did not return an extension module

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.6/site.py", line 544, in <module>
    main()
  File "/usr/lib/python3.6/site.py", line 531, in main
    known_paths = addsitepackages(known_paths)
  File "/usr/lib/python3.6/site.py", line 328, in addsitepackages
    addsitedir(sitedir, known_paths)
  File "/usr/lib/python3.6/site.py", line 207, in addsitedir
    addpackage(sitedir, name, known_paths)
  File "/usr/lib/python3.6/site.py", line 178, in addpackage
    import traceback
[...]

PR #190, opened 2018-08-22
Support virtualenv/venv without separate cache

tkf:
(To be rebased after #188)

With this patch, we should be able to have multiple Python virtual environments without duplicating Julia's precompilation cache, provided that those Python executables are linked against same libpython.

Since PyCall.jl only cares about the identity of libpython when it is already initialized https://github.com/JuliaPy/pyjulia/issues/182#issuecomment-413341399, i.e., it does [not](https://github.com/JuliaPy/PyCall.jl/blob/f1a63572df9ebe7c8908674e6ade596de495fe73/src/pyinit.jl#L79-L89) call `Py_SetPythonHome` etc., I think we only need to compare the path to libpython.

@stevengj Is it correct?  It seems that it works in my laptop.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,10 +13,15 @@ environment:
     BATDIR: ci\appveyor\win32
     CROSS_VERSION: 1
 
-  # 32 julia latest Python-35
-  - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
+  # 32 julia-1.0 Python-35
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/1.0/julia-1.0-latest-win32.exe"
     PYTHONDIR: "C:\\Python35"
     BATDIR: ci\appveyor\win32
+
+  # 32 julia latest Python-35
+  # - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
+  #   PYTHONDIR: "C:\\Python35"
+  #   BATDIR: ci\appveyor\win32
 
   # 64 julia-0.6 Python-35
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
@@ -24,10 +29,15 @@ environment:
     BATDIR: ci\appveyor\win64
     CROSS_VERSION: 1
 
-  # 64 julia latest Python-35
-  - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
+  # 64 julia-1.0 Python-35
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/1.0/julia-1.0-latest-win64.exe"
     PYTHONDIR: "C:\\Python35-x64"
     BATDIR: ci\appveyor\win64
+
+  # 64 julia latest Python-35
+  # - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
+  #   PYTHONDIR: "C:\\Python35-x64"
+  #   BATDIR: ci\appveyor\win64
 
 matrix:
   allow_failures:
--- a/julia/core.py
+++ b/julia/core.py
@@ -264,9 +264,7 @@ JuliaInfo = namedtuple(
     'JuliaInfo',
     ['JULIA_HOME', 'libjulia_path', 'image_file',
      # Variables in PyCall/deps/deps.jl:
-     'pyprogramname', 'libpython'],
-    # PyCall/deps/deps.jl may not exist; The variables are then set to None:
-    defaults=[None, None])
+     'pyprogramname', 'libpython'])
 
 
 def juliainfo(runtime='julia'):
@@ -297,6 +295,7 @@ def juliainfo(runtime='julia'):
         # object file: No such file or directory":
         env=_enviorn)
     args = output.decode("utf-8").rstrip().split("\n")
+    args.extend([None] * (len(JuliaInfo._fields) - len(args)))
     return JuliaInfo(*args)
 
 
--- a/julia/core.py
+++ b/julia/core.py
@@ -42,6 +42,12 @@ if python_version.major == 3:
 else:
     iteritems = dict.iteritems
 
+
+# As setting up Julia modifies os.environ, we need to cache it for
+# launching subprocesses later in the original environment.
+_enviorn = os.environ.copy()
+
+
 class JuliaError(Exception):
     pass
 
@@ -278,7 +284,11 @@ def juliainfo(runtime='julia'):
              include(PyCall_depsfile)
              println(pyprogramname)
          end
-         """])
+         """],
+        # Use the original environment variables to avoid a cryptic
+        # error "fake-julia/../lib/julia/sys.so: cannot open shared
+        # object file: No such file or directory":
+        env=_enviorn)
     args = output.decode("utf-8").rstrip().split("\n")
     if len(args) == 3:
         args.append(None)  # no pyprogramname set
--- a/julia/core.py
+++ b/julia/core.py
@@ -32,6 +32,8 @@ from ctypes import py_object
 # this is python 3.3 specific
 from types import ModuleType, FunctionType
 
+from .find_libpython import find_libpython, normalize_path
+
 #-----------------------------------------------------------------------------
 # Classes and funtions
 #-----------------------------------------------------------------------------
@@ -260,7 +262,11 @@ def determine_if_statically_linked():
 
 JuliaInfo = namedtuple(
     'JuliaInfo',
-    ['JULIA_HOME', 'libjulia_path', 'image_file', 'pyprogramname'])
+    ['JULIA_HOME', 'libjulia_path', 'image_file',
+     # Variables in PyCall/deps/deps.jl:
+     'pyprogramname', 'libpython'],
+    # PyCall/deps/deps.jl may not exist; The variables are then set to None:
+    defaults=[None, None])
 
 
 def juliainfo(runtime='julia'):
@@ -283,6 +289,7 @@ def juliainfo(runtime='julia'):
          if PyCall_depsfile !== nothing && isfile(PyCall_depsfile)
              include(PyCall_depsfile)
              println(pyprogramname)
+             println(libpython)
          end
          """],
         # Use the original environment variables to avoid a cryptic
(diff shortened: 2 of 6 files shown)


julia/core.py
```diff
@@ -443,26 +496,44 @@ def __init__(self, init_julia=True, jl_runtime_path=None, jl_init_path=None,
...
                 self._call(u"""
-                    isdir(Base.LOAD_CACHE_PATH[1]) ||
-                        mkpath(Base.LOAD_CACHE_PATH[1])
-                    depsfile = joinpath(Base.LOAD_CACHE_PATH[1],"PyCall.ji")
-                    isfile(depsfile) || touch(depsfile)
+                for path in Base.LOAD_CACHE_PATH[2:end]
+                    cache = joinpath(path, "PyCall.ji")
+                    backup = joinpath(path, "PyCall.ji.backup")
+                    if isfile(cache)
+                        mv(cache, backup; remove_destination=true)
```
tkf: This is a bit orthogonal to this PR but I figured fixing #175 is required for passing the test.  Merging this fix should also resolve #109 and #169.

In `.travis.yml` we have:

https://github.com/JuliaPy/pyjulia/blob/138d47dce6d6189be9ab6a8ea330cec3a237a8be/.travis.yml#L62-L64

Previously, `use_separate_cache` was `True` for both `py` and `py27` virtual environments.  Thus, no global cache `~/.julia/lib/v0.6/PyCall.ji` existed while running the above test.  However, with this patch, `use_separate_cache` is `False` in `py` because the code I added sees that `PYTHON` configured for `PyCall` uses the same `libpython` used by `py` virtual environment prepared by `tox`.  So, after running `tox -e py`, there is a file `~/.julia/lib/v0.6/PyCall.ji`.  Now then running `tox -e py27` hits the bug #175 because `julia` 0.6 does not recompile `PyCall` when it finds `~/.julia/lib/v0.6/PyCall.ji` in `Base.LOAD_CACHE_PATH[2]`.

julia/core.py
```diff
@@ -243,13 +249,45 @@ def isafunction(julia, julia_name, mod_name=""):
...
+    -------
+    libpython : str or None
+        A path to libpython or None if it's not found (e.g., when
+        statically linked).
+    """
+    if sys.platform.startswith('linux'):
+        return libpython_from_ldd_output(subprocess.check_output(
+            ["ldd", path],
+            universal_newlines=True))
+# TODO: somebody has to write it for in Windows and macOS:
```
tkf: Note: it only works in Linux at the moment

julia/core.py
```diff
@@ -243,13 +249,46 @@ def isafunction(julia, julia_name, mod_name=""):
...
+    libpython : str or None
+        A path to libpython or None if it's not found (e.g., when
+        statically linked).
+    """
+    if sys.platform.startswith('linux'):
+        return libpython_from_ldd_output(subprocess.check_output(
+            ["ldd", path],
+            universal_newlines=True))
+    return None
+# TODO: somebody has to write it for Windows and macOS:
```
tkf: Note: it only works in Linux at the moment

tkf:
I merged #188 and rebased this PR onto master.  There are test failures in macOS. But they are likely to be due to a bug in the latest pytest https://github.com/pytest-dev/pytest/issues/3888 (Those failures are coming from doctests.  Functions tested there are pure Python "pure" functions.  So it doesn't make sense to have different results in different OSes.)

@stevengj Can you have a look at it? Also please let me know if the idea of checking only the libpython path (my reasoning is in the issue description https://github.com/JuliaPy/pyjulia/pull/190#issue-210024732) is the correct direction.

stevengj:
It's annoying that finding `libpython` is so hard.  See [here](https://github.com/JuliaPy/PyCall.jl/blob/master/deps/build.jl#L50-L135) for how PyCall does it.

The analogue of `ldd` on MacOS is `otool -L`, but that's [only installed](https://apple.stackexchange.com/questions/28110/is-otool-available-in-a-stock-os-x-installation-with-no-xcode) if the user installs the optional developer tools.

So, I think that we should fall back to searching the `distutils.sysconfig.get_config_var('...')` paths as PyCall does.

ihnorton:
> The analogue of ldd on MacOS is otool -L, but that's only installed if the user installs the optional developer tools.

FWIW, there is a pure-Julia library which can do this: https://github.com/staticfloat/ObjectFile.jl (see `find_libraries`).

tkf:
@ihnorton Thanks. It looks like a good solution. Maybe PyCall.jl can use it in the build script and also expose it as a script tool. Then PyJulia can just call it.

But I've already ported `find_libpython` to a Python function :)  I was actually about to suggest to use the script `find_libpython.py` I added from PyCall.jl as well since there is no point in repeating this and doing from Python is slightly easier.  But if there is an out-of-the-box solution in Julia why don't we use it.

tkf:
Actually, maybe we should just fetch `dli_fbase` as done in `PyCall/src/startup.jl`?
https://github.com/JuliaPy/PyCall.jl/blob/9224c606545d7659ab89d8389ac9f44b1f45d1f9/src/startup.jl#L64-L77

appveyor.yml
```diff
@@ -13,22 +13,32 @@ environment:
     BATDIR: ci\appveyor\win32
     CROSS_VERSION: 1
 
-  # 32 julia latest Python-35
-  - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
+  # 32 julia-1.0 Python-35
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/1.0/julia-1.0-latest-win32.exe"
```
tkf: Testing with Julia nightly yields some bizarre failures (see below).  The same code works with Julia 1.0.  I suggest to test with stable release.

From Python 3.5:


> `ErrorException("ccall: could not find function PyString_AsStringAndSize in library C:\projects\pyjulia\.tox\py\Scripts\python35")`
> --- https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.175/job/wwr318to7jr1b5h6#L433

From Python 2.7:

> ```
>     def test_call_julia_function_with_python_args(self):
>         self.assertEqual(['A', 'B', 'C'],
>                          list(julia.map(julia.uppercase,
> >                                       array.array('u', [u'a', u'b', u'c']))))
> E       RuntimeError: Julia exception: MethodError(uppercase, (PyObject u'a',), 0x000061ca)
> ```
> --- https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.175/job/wwr318to7jr1b5h6#L547

julia/core.py
```diff
@@ -277,24 +287,61 @@ def juliainfo(runtime='julia'):
...
+    if determine_if_statically_linked():
+        _debug(sys.executable, "is statically linked.")
+        return False
+
+    # Note that the following check is OK since statically linked case
+    # is already excluded.
+    if is_same_path(jlinfo.pyprogramname, sys.executable):
+        # In macOS and Windows, find_libpython does not work as good
+        # as in Linux.  We add this shortcut so that PyJulia can work
+        # in those environments.
```
tkf: I needed this shortcut to make it work since `find_libpython.py` and `PyCall/deps/build.jl` sometimes disagree:


For example, in macOS, `PyCall/deps/build.jl` finds:

> [L957](https://travis-ci.org/JuliaPy/pyjulia/jobs/421978787#L957) `jlinfo.libpython = /usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/libpython3.7m`

but `find_libpython.py` finds:

> [L958](https://travis-ci.org/JuliaPy/pyjulia/jobs/421978787#L958) `py_libpython = /usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/Python`


Another example in Windows: `PyCall/deps/build.jl` cannot find a fullpath:

> [L517](https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.174/job/gj7k6da28y0wq3rd#L517) `jlinfo.libpython = python27`

but `find_libpython.py` "finds" it:

> [L518](https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.174/job/gj7k6da28y0wq3rd#L518) `py_libpython = C:\Windows\system32\python27.dl`

(But it looks like the same path was found in win32 and win64.  Ultimately, we need someone to write a function that uses `dladdr` equivalent in Windows.  It should be possible by just extending: https://stackoverflow.com/a/16659821)


I think it's better to use the same script to discover libpython in PyCall.jl and PyJulia.  We can then avoid a hack like this.  Since `find_libpython.py` can call `libdl.dladdr` (though it works only if libpython is dynamically linked), I think it's better approach.

stevengj: In the cases where they disagree, which is correct?  Should deps/build.jl be fixed?

tkf: I don't have access to macOS so I'm not sure. But PyJulia printed

```
jlinfo.libpython = /usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/libpython3.7m
py_libpython = /usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/Python
jl_libpython = None
```

where `jlinfo.libpython` is `libpython` saved in `deps/deps.jl` and `jl_libpython` is the path I tried to resolved by

https://github.com/JuliaPy/pyjulia/blob/6982ae885989285fc58c4b5181071b79975ea7f7/julia/find_libpython.py#L176-L182

So `jl_libpython = None` means that `libpython3.7m.dylib` did not exist (I set `suffix = ".dylib"` in macOS).  Maybe we should try `libpython3.7m.so` in macOS too?

In the case of Windows, `deps/build.jl` just gave up (`jlinfo.libpython = python27`) whereas `ctypes.util.find_library` finds `C:\Windows\system32\python27.dl`.

tkf: I'm trying it: 1d599be8d1d2e3dbf68b18ce07afc1afeea6ed51

tkf:
Re: https://github.com/JuliaPy/pyjulia/pull/190#discussion_r213847093

So with some more manual handling in macOS cdbefb19453e29c6b64f9c4b96857c4d441bb171 now `normalize_path` can normalize what `deps/build.jl` finds to what `find_libpython.py` finds.

```
jlinfo.libpython = /usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/libpython3.7m
py_libpython = /usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/Python
jl_libpython = /usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/Python
use_separate_cache = False
```
--- https://travis-ci.org/JuliaPy/pyjulia/jobs/422307471#L954

julia/find_libpython.py
```diff
@@ -0,0 +1,279 @@
...
+SHLIB_SUFFIX = sysconfig.get_config_var("SHLIB_SUFFIX")
+if SHLIB_SUFFIX is None:
+    if is_windows:
+        SHLIB_SUFFIX = ".dll"
+    else:
+        SHLIB_SUFFIX = ".so"
+if is_apple:
+    # sysconfig.get_config_var("SHLIB_SUFFIX") can be ".so" in macOS.
+    # Let's not use the value from sysconfig.
+    SHLIB_SUFFIX = ".dylib"
```
tkf: Here is the output of `python -m sysconfig` from Python 3.7 in macOS https://travis-ci.org/JuliaPy/pyjulia/jobs/421924267#L1439

tkf:
So to directly answer https://github.com/JuliaPy/pyjulia/pull/190#discussion_r213841789 `deps/build.jl` and `find_libpython.py` agree in all cases where `deps/build.jl` finds the path.

That is to say, they both agree in Linux and macOS.  In Windows and Python 2.7, `deps/build.jl` does not find libpython but `find_libpython.py` finds it via `ctypes.util.find_library`.  Presumably, this is equivalent to what `dlopen` does.  Python document says:

> The purpose of the `find_library()` function is to locate a library in a way similar to what the compiler or runtime loader does (on platforms with several versions of a shared library the most recent should be loaded), while the ctypes library loaders act like when a program is run, and call the runtime loader directly.
>
> --- https://docs.python.org/3/library/ctypes.html#finding-shared-libraries

stevengj:
Sounds like deps/build.jl should be updated, then.

tkf:
Why don't we use `find_libpython.py` in `PyCall.jl` as well?  Since we can use `dladdr`/`GetModuleFileName` inside a Python process (if libpython is dynamically linked), I think this is much more robust solution.

PR #193, opened 2018-08-24
Add IPython completion hook in julia.magic

tkf:
(To be rebased after #190)

image

### DIFF
--- a/julia/magic.py
+++ b/julia/magic.py
@@ -66,6 +66,74 @@ class JuliaMagics(Magics):
         return ans
 
 
+class JuliaCompleter(object):
+
+    """
+    Simple completion for ``%julia`` line magic.
+    """
+
+    @property
+    def jlcomplete_texts(self):
+        try:
+            return self._jlcomplete_texts
+        except AttributeError:
+            pass
+
+        julia = Julia()
+        if julia.eval('VERSION < v"0.7-"'):
+            self._jlcomplete_texts = lambda *_: []
+            return self._jlcomplete_texts
+
+        self._jlcomplete_texts = julia.eval("""
+        import REPL
+        (str, pos) -> begin
+            ret, _, should_complete =
+                REPL.completions(str, pos)
+            if should_complete
+                return map(REPL.completion_text, ret)
+            else
+                return []
+            end
+        end
+        """)
+        return self._jlcomplete_texts
+
+    def complete_command(self, _ip, event):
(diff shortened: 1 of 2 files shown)


tkf:
...or I can monkey-patch IPython :stuck_out_tongue:  https://github.com/tkf/ipyjulia_hacks

julia/magic.py
```diff
@@ -80,3 +111,9 @@ def julia(self, line, cell=None):
 def load_ipython_extension(ip):
     """Load the extension in IPython."""
     ip.register_magics(JuliaMagics)
+    ip.set_hook("complete_command", julia_completer,
+                re_key="%?%julia")
```
tkf: As it does not work with cell magic, just use `str_key="%julia"`.

julia/magic.py
```diff
@@ -66,6 +66,37 @@ def julia(self, line, cell=None):
...
+    (string, pos) -> begin
+        ret, _, should_complete =
+            REPL.completions(string, pos)
+        if should_complete
+            return map(REPL.completion_text, ret)
+        else
+            return String[]
+        end
+    end
+    """)(julia_code, julia_pos)
```
stevengj: should we `eval` this function once and cache it?

tkf: Yes. I use `cached_property` in my `ipyjulia_hacks`. Something like this can be done here. https://github.com/tkf/ipyjulia_hacks/blob/679f6bbd3ed202cde6ec6ebb4b1b537566531cdf/src/ipyjulia_hacks/completers.py#L19

But could you review #190 first? I think that's more fundamental and important. But that PR, in turn, depends on #188.

stevengj:
@Carreau, do you have a sense of what the best approach here is for cell magic?

Carreau:
Well, tab completion and monkey patching seem like a good approach for now. refactoring, improving  and getting for improving completion is in my goals.

I don't have anything short term solution though.

tkf:
@Carreau Can't wait to see https://github.com/ipython/ipython/pull/10722 (or similar mechna) arrive on IPython master!

@stevengj Do you want the monkey-patch in this PR?  This is how I did it:
https://github.com/tkf/ipyjulia_hacks/blob/579f43fb260e0c3e66f6e7b6f059aebe29a56c19/src/ipyjulia_hacks/ipy/completers.py#L89-L98

PR #197, opened 2018-09-05
Improve juliainfo

tkf:
Improvements:

* The original environment variables are used when launching `juliainfo`.  This makes it possible to call `juliainfo` after setting up `core.Julia`.  This should not happen when using `core.Julia` normally.  However, some bugs may invoke `juliainfo` after libjulia initialization.  This PR does not fix such bug but at least make the error message less insane.

* Make `juliainfo` callable even when PyCall.jl is not installed.  This makes the failure less confusing; i.e., PyJulia fails when `using PyCall` is evaluated.

* The standard error of julia is captured.  It is bundled in CalledProcessError in case of non-zero exit code.  In case of zero exit code, stderr emits warning if some message was written by julia.

* Currently, startup file is not loaded in the main PyJulia session. Including the startup file has some potential problem.  For example, if something is printed in the startup file then `juliainfo` would break.  So it seems to be that the safe default is to avoid loading the startup file (though it makes sense to add an option to load it in the future).

### DIFF
--- a/julia/core.py
+++ b/julia/core.py
@@ -274,7 +274,7 @@ def juliainfo(runtime='julia', **popen_kwargs):
     popen_kwargs.setdefault("env", _enviorn)
 
     proc = subprocess.Popen(
-        [runtime, "-e",
+        [runtime, "--startup-file=no", "-e",
          """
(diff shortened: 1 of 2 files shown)


tkf:
Rebased and fixed some issues. CIs are all green. It's good to go now.

PR #200, opened 2018-09-13
Add python-jl CLI

tkf:
This PR adds a CLI `python-jl` which acts like `python` CLI but executed as a `julia` process. This way, PyCall can use normal precompilation cache so that PyJulia can be used with Python installed in Debian/Ubuntu and conda #185.

The users just have to use `python-jl script.py` instead of `python script.py`.

(previously: https://github.com/JuliaPy/PyCall.jl/pull/562)

### DIFF
--- a/julia/core.py
+++ b/julia/core.py
@@ -399,7 +399,14 @@ For more information, see:
 _separate_cache_error_statically_linked = """
 Your Python interpreter "{sys.executable}"
 is statically linked to libpython.  Currently, PyJulia does not support
-such Python interpreter.  For available workarounds, see:
+such Python interpreter.  One easy workaround is to run your Python
+script with `python-jl` command bundled in PyJulia.  You can simply do:
+
+    python-jl PATH/TO/YOUR/SCRIPT.py
+
+See `python-jl --help` for more information.
+
+For other available workarounds, see:
     https://github.com/JuliaPy/pyjulia/issues/185
 """
 
--- a/julia/core.py
+++ b/julia/core.py
@@ -686,6 +686,10 @@ class Julia(object):
         self.sprint = self.eval('sprint')
         self.showerror = self.eval('showerror')
 
+        if self.eval('VERSION >= v"0.7-"'):
+            self.eval("@eval Main import Base.MainInclude: eval, include")
+            # https://github.com/JuliaLang/julia/issues/28825
+
     def _debug(self, *msg):
         """
         Print some debugging stuff, if enabled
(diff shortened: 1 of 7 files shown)


tkf:
In Windows, Launching `python-jl` fails due to `ERROR: syntax: incomplete: premature end of input`:

https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.203/job/kgic6eicchk4chx7#L776

Replacing `\n` with `\r\n` 7c395e3bd9e82387e339d7fae4595d851463c211 ([appveyor](https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.202)) and `\n` with `;` 0a61076ec0df05d5160b634b3458481ab1ff7707 ([appveyor](https://ci.appveyor.com/project/Keno/pyjulia/build/1.0.203)) didn't help.

Can we _not_ support `python-jl` in Windows?  In Windows Python is dynamically linked so no one has to use `python-jl` anyway.

tkf:
There is a bit of limitation in `argparse` (and `optparse`) for mimicking the behavior of `python` CLI.  Namely, it's hard to tell the parser to stop parsing (even with `argparse.REMAINDER` and `parse_known_args`).  I tried to do a simple workaround by pre-processing the arguments 4bcf10991b5ee9f82754bcd457bd8143601b83c0 but it didn't work in all the cases.  So I gave up and cooked up a small CLI parser.  Now it works identical to `python` CLI for all the cases I've tried.

Interestingly, `python-jl` can now be used to launch IPython kernel (to be used from Jupyter notebook/qtconsole/console).  You only need to make (say) `~/.local/share/jupyter/kernels/python-jl/kernel.json` with:

```json
{
  "argv": [
    "/PATH/TO/bin/python-jl",
    "-m",
    "ipykernel_launcher",
    "-f",
    "{connection_file}"
  ],
  "display_name": "Python.jl",
  "language": "python"
}
```

Note that all I needed to do was to replace `python` in normal `kernel.json` generated by `python -m ipykernel install --user --name python-jl --display-name "Python.jl"` with `python-jl`.

Everything should work almost like normal IPython kernel.  The only way to know that you are in Julia process is to do something like:

```
import os
!ps -f {os.getpid()}
```

which should show `julia -e ...`.

julia/core.py
```diff
@@ -650,6 +657,10 @@ def __init__(self, init_julia=True, jl_init_path=None, runtime=None,
         self.sprint = self.eval('sprint')
         self.showerror = self.eval('showerror')
 
+        if self.eval('VERSION >= v"0.7-"'):
+            self.eval("@eval Main import Base.MainInclude: eval, include")
+            # https://github.com/JuliaLang/julia/issues/28825
```
tkf: This should fix #202. The reason why it's in this PR is that this fix depends on https://github.com/JuliaPy/pyjulia/pull/200/commits/b908e6bd85664e9b846a80c58710fecb1028d822.

PR #201, opened 2018-09-14
Re-import PyCall in case init_julia=False is passed

tkf:
It fixes the following scenario:

```julia
julia> m = Module();

julia> Base.eval(m, quote
       using PyCall
       end)

julia> Base.eval(m, quote
       py"""
       from julia import Julia
       Julia(init_julia=False)
       """
       end)
```

### DIFF
--- a/julia/core.py
+++ b/julia/core.py
@@ -645,6 +645,7 @@ class Julia(object):
                 end
                 """)
 
+            # This is mainly for initiating the precompilation:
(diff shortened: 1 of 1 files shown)


tkf:
It looks like the failure in appveyor is just a network issue https://ci.appveyor.com/project/Keno/pyjulia/builds/19740497/job/x0flvinvn88jnvuo#L51

Merging...

PR #209, opened 2018-10-24
Teak setup.py

tkf:
* Pass markdown directly to `long_description` and specify  `long_description_content_type`.  Don't use pandoc.
* Load version from `julia/release.py`.  This number can be accessed via `julia.__version__` at runtime now.
* Bump version number to 0.2.0.dev.
* Pass `zip_safe=False`.


tkf:
Uploaded to https://test.pypi.org/project/julia/

Looks like it's rendered nicely:

image

PR #210, opened 2018-10-25
Add more information in README.md

tkf:
Preview: https://github.com/tkf/pyjulia/blob/readme/README.md

TODO:

- [x] Explain the problem with statically linked executable and the workarounds
- [x] Link https://packaging.python.org/tutorials/
- [x] Recommend IPython 7.0
- [ ] Maybe mention `conda-forge`? https://github.com/JuliaPy/pyjulia/issues/185#issuecomment-433229472; Waiting for https://github.com/conda-forge/python-feedstock/issues/222
- [x] Mention `%load_ext julia.magic`!
- [x] Mention that libjulia swallows all SIGINT: #211
- [x] GIL
- [x] virtual environments / libpython


tkf:
@stevengj Can you have a look at updated README? Especially the notes on the pre-compilation cache? https://github.com/tkf/pyjulia/blob/readme/README.md#pre-compilation-mechanism-in-julia-10

tkf:
Here is how it'd be rendered in PyPI https://test.pypi.org/project/julia/0.2.0.dev1/

README.md
```diff
@@ -6,62 +6,75 @@ PyJulia
...
+**Note:** If you are not familiar with `pip` and have some troubles
+with the following installation steps, we recommend to go through
+[Tutorials in Python Packaging User Guide](https://packaging.python.org/tutorials/).
 
 To get released versions you can use:
 
-```
-pip install julia
+```console
+$ python3 -m pip install --user julia
```
tkf: I switched `pip` to `python3 -m pip` everywhere since it's a common pitfall that `pip` uses different Python executable than `python`.

tkf: ... and added `--user` since that's likely what they want.

stevengj:
Fine, although what we did previously was not technically monkey patching. I’m not sure it is a good idea to try to explain what is going wrong in the readme. Just say that pecompilation doesn’t work with 1.0 and link the issue.

tkf:
We are changing the path of the Julia executable which is conceptually like mutating a function.  But it is done through the public API (but in an implementation-dependent manner) so I guess I'm stretching it somewhat to call it a monkey patch.

The pre-compilation issue was asked by multiple users (including me) so I think it would be nice to explain it in some details.  How about blending it into `fake-julia/README`?

tkf:
Here is the new fake-julia documentation I suggest: https://github.com/tkf/pyjulia/tree/readme/julia/fake-julia

tkf:
@stevengj Thanks a lot for reviewing!

PR #213, opened 2018-10-26
add unexported variables to module's __dir__

mtikekar:
__dir__ should have names of all variables and not just the exported ones like in __all__

### DIFF
--- a/julia/core.py
+++ b/julia/core.py
@@ -96,7 +96,12 @@ class JuliaModule(ModuleType):
             names = set()
         else:
             names = set(super(JuliaModule, self).__dir__())
-        names.update(self.__all__)
+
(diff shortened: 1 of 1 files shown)


julia/core.py
```diff
@@ -96,7 +96,12 @@ def __dir__(self):
             names = set()
         else:
             names = set(super(JuliaModule, self).__dir__())
-        names.update(self.__all__)
+
+        juliapath = remove_prefix(self.__name__, "julia.")
+        julianames = set(self._julia.eval("names({}; all=true)".format(juliapath)))
```
tkf: `names({}; all=true)` does not work with Julia 0.6.

PR #230, opened 2019-01-03
Test with Julia 1.1

tkf:


tkf:
Yeah, maybe IPython changed version bound or something in their release to nudge people to switch to Python 3. I think it can be solved by bounding IPython version in our test setup (tox.ini).

tkf:
...although it's puzzling why it happened in just one build in windows.

tkf:
I didn't do anything but the old failure (https://ci.appveyor.com/project/Keno/pyjulia/builds/21347376) is now fixed (https://ci.appveyor.com/project/Keno/pyjulia/builds/22426825).  Maybe it was some upstream issue (in pip or setuptools?).

But I'm closing this ATM since it's better to add test with 1.1 once we drop 0.6, rather than replacing 1.0 with 1.1.

PR #247, opened 2019-03-13
Automatically use --compiled-modules=no in test

tkf:
The aim is to make it possible to run tests even with Python runtime incompatible with compiled PyCall cache.


tkf:
#261 is a better approach

PR #254, opened 2019-03-25
Load modules with correct name based on `fullname`

tkf:
Before this patch:

    In [1]: from julia import Main
       ...: Main.eval("import Base.Enums")
       ...: from julia.Base import Enums
       ...: Enums
    Out[1]: <PyCall.jlwrap Base.Enums>

After this patch, `Out[1]` is `<module 'julia.Base.Enums' ...>` as expected.

### DIFF
--- a/julia/pyjulia_helper.jl
+++ b/julia/pyjulia_helper.jl
@@ -1,5 +1,36 @@
 module _PyJuliaHelper
 
+if VERSION < v"0.7-"
+nameof(m::Module) = ccall(:jl_module_name, Ref{Symbol}, (Any,), m)
+
+parentmodule(m::Module) = ccall(:jl_module_parent, Ref{Module}, (Any,), m)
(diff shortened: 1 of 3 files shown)


julia/pyjulia_helper.jl
```diff
@@ -1,5 +1,36 @@
 module _PyJuliaHelper
 
+if VERSION < v"0.7-"
```
stevengj: Seems like pyjulia should drop support for pre-1.0 versions of Julia.

tkf: Yeah, I guess I'll do it after releasing 0.3.  It has no major blocker at the moment so it can happen anytime.  I thought I'd keep Julia 0.6 compatibility to make sure refactoring keeps the flexibility to support separated cache (which in retrospect didn't help much).

PR #258, opened 2019-03-27
Intuitive handling of sysimage relative path

tkf:
Julia tries to interpret `sysimage` as a path relative to `bindir` and aborts if not found.  With this PR, it is turned to absolute path before passed to Julia.

See also: https://github.com/JuliaLang/julia/pull/31492

### DIFF
--- a/src/julia/core.py
+++ b/src/julia/core.py
@@ -601,6 +601,25 @@ class LibJulia(BaseLibJulia):
     However, a proper use of the C-API is more involved and presumably
     very challenging without C macros.  See also:
     https://docs.julialang.org/en/latest/manual/embedding/
+
+    Attributes
+    ----------
+    libjulia_path : str
+        Path to libjulia.
+    bindir : str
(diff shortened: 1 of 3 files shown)


src/julia/core.py
```diff
@@ -601,6 +601,25 @@ class LibJulia(BaseLibJulia):
...
+    sysimage : str
+        Path to system image.  This is passed to `jl_init_with_image`
+        unless overridden by argument ``option`` to `init_julia`.
+
+        .. warning::
+
+            Unlike `jl_init_with_image` C API, relative path is not
+            interpreted as relative to `bindir`.  Instead, relative
+            path is resolved using `os.path.realpath` before passed to
+            Julia.
```
stevengj: I would simply say that if `sysimage` is a relative path, it is interpreted relative to the current directory (rather than relative to the Julia `bindir` in the `jl_init_with_image` C API).

Issue #205, opened 2018-10-07
Add new magic option that loads python variable into julia


clouds56:
usage:
```
x, y = 1, 2
```
and in anther shell
```
%%julia --input x,y
println(x+y)
```

clouds56:
That's great!
I didn't know this is implementable.

stevengj:
It's already supported by PyCall in `py"...."` strings, so this would be similar.

tkf:
(I was meant to write this for #262 but I figured here is a better place for this.)

I think we need to write own parer like PyCall does to make it feel really "natural."  For the moment, maybe we can have `--format` (`-f`) option to `%%julia` to use the `str.format` mini-language such that something like the following work

```
%%julia --format
data = {x}    # the data of Python variable x is transferred
code = {x!c}  # x is interpreted as Julia code
str = {x!s}   # x is interpreted as a string literal in Julia
num = {x:.2}  # x is interpreted as a numeric literal in Julia (with given decimal precision)
TT = Tuple{{Any,Any}}  # {} has to be escaped manually
```

This is not as elegant as `$(python code)` but it's good to reuse the concept that Python programmers already know.

I think the custom conversion specifier `c` can be implemented through [`object.__format__`](https://docs.python.org/3/reference/datamodel.html#object.__format__) protocol.  But I haven't checked it yet.

tkf:
Actually, I just remembered that `py"..."` macro already works for doing something similar

```python
In [1]: %load_ext julia.magic
Initializing Julia interpreter. This may take some time...

In [2]: x = 1

In [3]: %%julia
   ...: using PyCall
   ...: py"x + 1"
Out[3]: 2
```

PR #262, opened 2019-04-01
add variable interpolation

marius311:
From the docs included in this PR:

Python variables can be passed to Julia using either `$foo` or `{foo}`.  

```none
In [3]: foo = [[1,2],[3,4]]

In [4]: %julia $foo .+ 1
Out[4]: 
array([[2, 3],
       [4, 5]], dtype=int64)
```

The `{}` style supports passing arbitrary Python expressions inside.

```none
In [5]: %julia {[x for x in range(3)]}
Out[5]: array([0, 1, 2], dtype=int64)
```

Variables are automatically converted between equivalent Python/Julia types (should they exist) using PyCall. Note below that `1`, `x`, and the tuple itself are converted to Julia `Int64`, `String`, and `Tuple`, respectively, although the function `abs` and the result of `typeof` are not.

```none
In [6]: %julia typeof({(1,"x",abs)})
Out[6]: <PyCall.jlwrap Tuple{Int64,String,PyObject}>
```

Closes https://github.com/JuliaPy/pyjulia/issues/205

Happy to take any feedback or requests for changes. I didn't write any tests since I couldn't figure out how to make pytest work with python-jl (which my system needs) so maybe someone could help with creating tests, perhaps in a separate PR? (there we'll need to figure out how testing interacts with the fact that interpolation is using the IPython shell global namespace)

### DIFF
--- a/docs/source/usage.md
+++ b/docs/source/usage.md
@@ -70,7 +70,7 @@ You can then use, e.g.,
 In IPython (and therefore in Jupyter), you can directly execute Julia
 code using `%%julia` magic:
 
-```
+```none
 In [1]: %load_ext julia.magic
 Initializing Julia interpreter. This may take some time...
 
@@ -84,6 +84,35 @@ In [2]: %%julia
   | | |_| | | | (_| |  |  Version 1.0.1 (2018-09-29)
  _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
 |__/                   |
+```
+
+Any `$var` or `$(var)` expressions that appear in the Julia code (except in
+comments or string literals) are evaluated in Python and passed to Julia. Use
+`$$var` to insert a literal `$var` in Julia code.
+
+```none
+In [3]: foo = [[1,2],[3,4]]
+
+In [4]: %julia $foo .+ 1
+
+Out[4]: 
+array([[2, 3],
+       [4, 5]], dtype=int64)
+       
+In [22]: %%julia
+   ...: bar=3
+   ...: :($$bar)
+   
+Out[22]: 3
+
+```
+
+Variables are automatically converted between equivalent Python/Julia types (should they exist) using PyCall. Note below that `1`, `x`, and the tuple itself are converted to Julia `Int64`, `String`, and `Tuple`, respectively, although the function `abs` and the result of `typeof` are not.
+
+```none
+In [6]: %julia typeof($(1,"x",abs))
+
+Out[6]: <PyCall.jlwrap Tuple{Int64,String,PyObject}>
 ```
 
 #### IPython configuration
--- a/docs/source/usage.md
+++ b/docs/source/usage.md
@@ -91,15 +91,22 @@ In [5]: %julia sum(py"[x**2 for x in arr]")
 Out[5]: 14
 ```
 
-Python interpolation is not performed inside of strings (instead this is treated as regular Julia string interpolation), and can also be overridden elsewhere by using `$$...` to insert a literal `$...`:
+Interpolation from Python is only done outside of strings and quote blocks; to interpolate into these, you'll need to "escape" them one extra time:
 
 ```julia
-In [6]: %julia foo=3; "$foo"
-Out[6]: '3'
-
-In [7]: %julia bar=3; :($$bar)
-Out[7]: 3
+In [6]: foo="Python"
+        %julia foo="Julia"
+        %julia "this is $foo", "this is $($foo)"
+Out[6]: ('this is Julia', 'this is Python')
 ```
+
+Expressions inside of macros are never interpolated from Python:
+
+```julia
+In [7]: %julia @eval $foo
+Out[7]: 'Julia'
+```
+
 
 Variables are automatically converted between equivalent Python/Julia types (should they exist). You can turn this off by appending `o` to the Python string:
 
--- a/docs/source/usage.md
+++ b/docs/source/usage.md
@@ -79,9 +79,10 @@ array([[2, 3],
        [4, 5]], dtype=int64)
 ```
 
-You can "interpolate" Python objects into Julia code via `$var` for the value of single variables or `py"expression"` for the result of any arbitrary Python code:
+You can call Python code from inside of `%julia` blocks via `$var` for accessing single variables or `py"..."` for more complex expressions:
+
 ```julia
-In [3]: arr = [1,2,3]
+In [3]: arr = [1, 2, 3]
 
 In [4]: %julia $arr .+ 1
 Out[4]: 
@@ -91,31 +92,30 @@ In [5]: %julia sum(py"[x**2 for x in arr]")
 Out[5]: 14
 ```
 
-Interpolation from Python is only done outside of strings and quote blocks; to interpolate into these, you'll need to "escape" them one extra time:
+Inside of strings and quote blocks, `$var` and `py"..."` don't call Python and instead retain their usual Julia behavior. To call Python code in these cases, you can "escape" one extra time:
 
 ```julia
-In [6]: foo="Python"
-        %julia foo="Julia"
-        %julia "this is $foo", "this is $($foo)"
+In [6]: foo = "Python"
+        %julia foo = "Julia"
+        %julia ("this is $foo", "this is $($foo)")
 Out[6]: ('this is Julia', 'this is Python')
 ```
 
-Expressions inside of macros are never interpolated from Python:
+Expressions in macro arguments also always retain the Julia behavior:
 
 ```julia
 In [7]: %julia @eval $foo
 Out[7]: 'Julia'
 ```
 
-
-Variables are automatically converted between equivalent Python/Julia types (should they exist). You can turn this off by appending `o` to the Python string:
+Results are automatically converted between equivalent Python/Julia types (should they exist). You can turn this off by appending `o` to the Python string:
 
 ```python
 In [8]: %julia typeof(py"1"), typeof(py"1"o)
 Out[8]: (<PyCall.jlwrap Int64>, <PyCall.jlwrap PyObject>)
 ```
 
-Interpolated variables obey Python scope, as expected:
+Code inside `%julia` blocks obeys the Python scope:
 
 ```python
 In [9]: x = "global"
--- a/docs/source/usage.md
+++ b/docs/source/usage.md
@@ -67,7 +67,7 @@ You can then use, e.g.,
 
 ### IPython magic
 
-In IPython (and therefore in Jupyter), you can directly execute Julia code using `%%julia` magic:
+In IPython (and therefore in Jupyter), you can directly execute Julia code using `%julia` magic:
 
 ```python
 In [1]: %load_ext julia.magic
@@ -79,8 +79,7 @@ array([[2, 3],
        [4, 5]], dtype=int64)
 ```
 
-You can "interpolate" Python results into Julia code via `$var` for single variable names, `$(expression)` for *most* Python code (although notably excluding comprehensions and any Python syntax which is not also valid Julia syntax), or `py"expression"` for *any* arbitrary Python code:
-
+You can "interpolate" Python objects into Julia code via `$var` for the value of single variables or `py"expression"` for the result of any arbitrary Python code:
 ```julia
 In [3]: arr = [1,2,3]
 
@@ -88,40 +87,37 @@ In [4]: %julia $arr .+ 1
 Out[4]: 
 array([2, 3, 4], dtype=int64)
 
-In [5]: %julia $(len(arr))
-Out[5]: 3
-
-In [6]: %julia py"[x**2 for x in arr]"
-Out[6]: array([1, 4, 9], dtype=int64)
+In [5]: %julia sum(py"[x**2 for x in arr]")
+Out[5]: 14
 ```
 
-Interpolation is never performed inside of strings. If you wish to override interpolation elsewhere, use `$$...` to insert a literal `$...`:
+Python interpolation is not performed inside of strings (instead this is treated as regular Julia string interpolation), and can also be overridden elsewhere by using `$$...` to insert a literal `$...`:
 
 ```julia
-In [7]: %julia foo=3; "$foo"
-Out[7]: '3'
+In [6]: %julia foo=3; "$foo"
+Out[6]: '3'
 
-In [8]: %julia bar=3; :($$bar)
-Out[8]: 3
+In [7]: %julia bar=3; :($$bar)
+Out[7]: 3
 ```
 
 Variables are automatically converted between equivalent Python/Julia types (should they exist). You can turn this off by appending `o` to the Python string:
 
 ```python
-In [9]: %julia typeof(py"1"), typeof(py"1"o)
-Out[9]: (<PyCall.jlwrap Int64>, <PyCall.jlwrap PyObject>)
+In [8]: %julia typeof(py"1"), typeof(py"1"o)
+Out[8]: (<PyCall.jlwrap Int64>, <PyCall.jlwrap PyObject>)
 ```
 
-Note that interpolated variables always refer to the global Python scope:
+Interpolated variables obey Python scope, as expected:
 
 ```python
-In [10]: x = "global"
-    ...: def f():
-    ...:     x = "local"
-    ...:     ret = %julia py"x"
-    ...:     return ret
-    ...: f()
-Out[10]: 'global'
+In [9]: x = "global"
+   ...: def f():
+   ...:     x = "local"
+   ...:     ret = %julia py"x"
+   ...:     return ret
+   ...: f()
+Out[9]: 'local'
 ```
 
 #### IPython configuration
--- a/docs/source/usage.md
+++ b/docs/source/usage.md
@@ -67,52 +67,61 @@ You can then use, e.g.,
 
 ### IPython magic
 
-In IPython (and therefore in Jupyter), you can directly execute Julia
-code using `%%julia` magic:
+In IPython (and therefore in Jupyter), you can directly execute Julia code using `%%julia` magic:
 
-```none
+```python
 In [1]: %load_ext julia.magic
 Initializing Julia interpreter. This may take some time...
 
-In [2]: %%julia
-   ...: Base.banner(IOContext(stdout, :color=>true))
-               _
-   _       _ _(_)_     |  Documentation: https://docs.julialang.org
-  (_)     | (_) (_)    |
-   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
-  | | | | | | |/ _` |  |
-  | | |_| | | | (_| |  |  Version 1.0.1 (2018-09-29)
- _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
-|__/                   |
-```
-
-Any `$var` or `$(var)` expressions that appear in the Julia code (except in
-comments or string literals) are evaluated in Python and passed to Julia. Use
-`$$var` to insert a literal `$var` in Julia code.
-
-```none
-In [3]: foo = [[1,2],[3,4]]
-
-In [4]: %julia $foo .+ 1
-
-Out[4]: 
+In [2]: %julia [1 2; 3 4] .+ 1 
+Out[2]: 
 array([[2, 3],
        [4, 5]], dtype=int64)
-       
-In [22]: %%julia
-   ...: bar=3
-   ...: :($$bar)
-   
-Out[22]: 3
-
 ```
 
-Variables are automatically converted between equivalent Python/Julia types (should they exist) using PyCall. Note below that `1`, `x`, and the tuple itself are converted to Julia `Int64`, `String`, and `Tuple`, respectively, although the function `abs` and the result of `typeof` are not.
+You can "interpolate" Python results into Julia code via `$var` for single variable names, `$(expression)` for *most* Python code (although notably excluding comprehensions and any Python syntax which is not also valid Julia syntax), or `py"expression"` for *any* arbitrary Python code:
 
-```none
-In [6]: %julia typeof($(1,"x",abs))
+```julia
+In [3]: arr = [1,2,3]
 
-Out[6]: <PyCall.jlwrap Tuple{Int64,String,PyObject}>
+In [4]: %julia $arr .+ 1
+Out[4]: 
+array([2, 3, 4], dtype=int64)
+
+In [5]: %julia $(len(arr))
+Out[5]: 3
+
+In [6]: %julia py"[x**2 for x in arr]"
+Out[6]: array([1, 4, 9], dtype=int64)
+```
+
+Interpolation is never performed inside of strings. If you wish to override interpolation elsewhere, use `$$...` to insert a literal `$...`:
+
+```julia
+In [7]: %julia foo=3; "$foo"
+Out[7]: '3'
+
+In [8]: %julia bar=3; :($$bar)
+Out[8]: 3
+```
+
+Variables are automatically converted between equivalent Python/Julia types (should they exist). You can turn this off by appending `o` to the Python string:
+
+```python
+In [9]: %julia typeof(py"1"), typeof(py"1"o)
+Out[9]: (<PyCall.jlwrap Int64>, <PyCall.jlwrap PyObject>)
+```
+
+Note that interpolated variables always refer to the global Python scope:
+
+```python
+In [10]: x = "global"
+    ...: def f():
+    ...:     x = "local"
+    ...:     ret = %julia py"x"
+    ...:     return ret
+    ...: f()
+Out[10]: 'global'
 ```
 
 #### IPython configuration
--- a/src/julia/magic.py
+++ b/src/julia/magic.py
@@ -104,9 +104,17 @@ class JuliaMagics(Magics):
         """
         src = compat.unicode_type(line if cell is None else cell)
 
+        # We assume the caller's frame is the first parent frame not in the
+        # IPython module. This seems to work with IPython back to ~v5, and
+        # is at least somewhat immune to future IPython internals changes,
+        # although by no means guaranteed to be perfect.
+        caller_frame = sys._getframe(3)
+        while caller_frame.f_globals.get('__name__').startswith("IPython"):
+            caller_frame = caller_frame.f_back
+        
         return self._julia.eval("""
         _PyJuliaHelper.@prepare_for_pyjulia_call begin %s end
-        """%src)(self.shell.user_ns, sys._getframe(4).f_locals)
+        """%src)(self.shell.user_ns, caller_frame.f_locals)
 
 # Add to the global docstring the class information.
 __doc__ = __doc__.format(
--- a/src/julia/magic.py
+++ b/src/julia/magic.py
@@ -22,7 +22,7 @@ from __future__ import print_function, absolute_import
 import sys
 import warnings
 
-from IPython.core.magic import Magics, magics_class, line_cell_magic
+from IPython.core.magic import Magics, magics_class, line_cell_magic, no_var_expand
 from IPython.utils import py3compat as compat
 from traitlets import Bool, Enum
 
@@ -89,6 +89,7 @@ class JuliaMagics(Magics):
         self._julia = Julia(init_julia=True)
         print()
 
+    @no_var_expand
     @line_cell_magic
     def julia(self, line, cell=None):
         """
@@ -97,14 +98,9 @@ class JuliaMagics(Magics):
         """
         src = compat.unicode_type(line if cell is None else cell)
 
-        try:
-            ans = self._julia.eval(src)
-        except JuliaError as e:
-            print(e, file=sys.stderr)
-            ans = None
-
-        return ans
-
+        return self._julia.eval("""
+        _PyJuliaHelper.@prepare_for_pyjulia_call begin %s end
+        """%src)(self.shell.user_ns, self.shell.user_ns)
 
 # Add to the global docstring the class information.
 __doc__ = __doc__.format(
--- a/src/julia/magic.py
+++ b/src/julia/magic.py
@@ -22,12 +22,18 @@ from __future__ import print_function, absolute_import
 import sys
 import warnings
 
-from IPython.core.magic import Magics, magics_class, line_cell_magic, no_var_expand
+from IPython.core.magic import Magics, magics_class, line_cell_magic
 from IPython.utils import py3compat as compat
 from traitlets import Bool, Enum
 
 from .core import Julia, JuliaError
 from .tools import redirect_output_streams
+
+try:
+    from IPython.core.magic import no_var_expand
+except ImportError:
+    def no_var_expand(f):
+        return f
 
 #-----------------------------------------------------------------------------
 # Main classes
--- a/src/julia/magic.py
+++ b/src/julia/magic.py
@@ -100,7 +100,7 @@ class JuliaMagics(Magics):
 
         return self._julia.eval("""
         _PyJuliaHelper.@prepare_for_pyjulia_call begin %s end
-        """%src)(self.shell.user_ns, self.shell.user_ns)
+        """%src)(self.shell.user_ns, sys._getframe(4).f_locals)
 
 # Add to the global docstring the class information.
 __doc__ = __doc__.format(
--- a/src/julia/pyjulia_helper.jl
+++ b/src/julia/pyjulia_helper.jl
@@ -49,48 +49,61 @@ if VERSION >= v"0.7-"
 end
 
 
+# takes an expression like `$foo + 1` and turns it into a pyfunction
+# `(globals,locals) -> convert(PyAny, pyeval_("foo",globals,locals,PyAny)) + 1`
+# so that Python code can all it and just pass the appropriate globals/locals
+# dicts to perform the interpolation. 
 macro prepare_for_pyjulia_call(ex)
     
-    # f(x) should return a transformed expression x and whether to recurse 
-    # into the new expression
-    function stoppable_walk(f, x)
-        (fx, recurse) = f(x)
-        walk(fx, (recurse ? (x -> stoppable_walk(f,x)) : identity), identity)
+    # f(x, quote_depth) should return a transformed expression x and whether to
+    # recurse into the new expression. quote_depth keeps track of how deep
+    # inside of nested quote objects we arepyeval
+    function stoppable_walk(f, x, quote_depth=1)
+        (fx, recurse) = f(x, quote_depth)
+        if isexpr(fx,:quote)
+            quote_depth += 1
+        end
+        if isexpr(fx,:$)
+            quote_depth -= 1
+        end
+        walk(fx, (recurse ? (x -> stoppable_walk(f,x,quote_depth)) : identity), identity)
     end
     
-    locals = gensym("locals")
-    globals = gensym("globals")
-    
-    function make_pyeval(expr::Union{String,Symbol}, options...)
+    function make_pyeval(globals, locals, expr::Union{String,Symbol}, options...)
         code = string(expr)
         T = length(options) == 1 && 'o' in options[1] ? PyObject : PyAny
         input_type = '\n' in code ? Py_file_input : Py_eval_input
-        :($convert($T, $pyeval_($code, $(Expr(:$,globals)), $(Expr(:$,locals)), $input_type)))
+        :($convert($T, $pyeval_($code, $globals, $locals, $input_type)))
     end
-        
-    ex = stoppable_walk(ex) do x
-        if isexpr(x, :$)
-            if isexpr(x.args[1], :$)
-                x.args[1], false
-            elseif x.args[1] isa Symbol
-                make_pyeval(x.args[1]), false
+    
+    function insert_pyevals(globals, locals, ex)
+        stoppable_walk(ex) do x, quote_depth
+            if quote_depth==1 && isexpr(x, :$)
+                if x.args[1] isa Symbol
+                    make_pyeval(globals, locals, x.args[1]), false
+                else
+                    error("""syntax error in: \$($(string(x.args[1])))
+                    Use py"..." instead of \$(...) for interpolating Python expressions.""")
+                end
+            elseif quote_depth==1 && isexpr(x, :macrocall)
+                if x.args[1]==Symbol("@py_str")
+                    # in Julia 0.7+, x.args[2] is a LineNumberNode, so filter it out
+                    # in a way that's compatible with Julia 0.6:
+                    make_pyeval(globals, locals, filter(s->(s isa String), x.args[2:end])...), false
+                else
+                    x, false
+                end
             else
-                error("""syntax error in: \$($(string(x.args[1])))
-                Use py"..." instead of \$(...) for interpolating Python expressions, 
-                or \$\$(...) for a literal Julia \$(...).
-                """)
+                x, true
             end
-        elseif isexpr(x, :macrocall) && x.args[1]==Symbol("@py_str")
-            # in Julia 0.7+, x.args[2] is a LineNumberNode, so filter it out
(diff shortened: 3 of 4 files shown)


stevengj:
Why `{}` instead of `$(...)`?

marius311:
> Why {} instead of $(...)?

It would require matching balanced parenthesis with regex (e.g. `$( 1 + (2+3) )`) which afaik is not possible. I guess I could do something that works to finite nesting, or go beyond regex, but it seemed too complex, whereas `$foo` and `{foo expression}` are already implemented in IPython in `DollarFormatter` hence why the code in this PR was quite simple. Up to you guys what you think should be done, I do agree ` $(...)` would be nice for consistency with Julia, but it might take me a while. 

I'll also work on fixing the test errors.

marius311:
Actually I realize now `{}` is not going to work either since its valid Julia code, so definitely something else needs to be done...

tkf:
> I realize now `{}` is not going to work either since its valid Julia code

I thought this PR was awesome until I saw this comment.  Too bad...

See also https://github.com/JuliaPy/pyjulia/issues/205#issuecomment-478755873

> how to make pytest work with python-jl

You can install pytest with `python-jl -m pip install pytest` and run it with, e.g., `python-jl -m pytest PATH/TO/TEST.py`.  Also, after #261 you should be able to do `tox -- --julia-compiled-modules=no` to run tests with statically linked Python.

marius311:
From @tkf's comment in the other thread. 
```
In [3]: %%julia
   ...: using PyCall
   ...: py"x + 1"
```
Ha, hadn't thought of that. Thinking about it more, in some sense that's the best syntactically-sound way to do this. 

The problem with either `$foo` or `{foo}` from my PR is both are sometimes valid Julia code. Consider e.g. 

```
%julia Vector{Int}
```

where `{Int}` not meant to be interpolated from Python, or even worse, 

```
x = 1

%%julia
y = $x
@eval $y
```
where `$x` is meant to be interpolated but `$y` is not. If you don't want to implement more complex expression parsing than regex, the way to catch these cases is use some characters other than `$` or `{}` that aren't valid Julia code. 

You could also in theory syntactically parse the expression, but actually that to me seems like exactly what using `py".."` does, its just that Julia itself is doing the parsing for you. So maybe that is the best solution?  

The limitation is that `py"..."` can only grab stuff in the global scope. Its also slightly wordy / unintuitive, so what about the following proposal for modifying this PR:
* I get rid of `{}` interpolation entirely. 
* I add a `%julia --raw` option which turns off interpolation, so you can do `%julia y=$x` then `%julia --raw @eval $y`. 
* We document that for expression substitutions, you can do `py"..."`. 

I gather this will work the way people expect it to, be flexible enough to let people do what they want by splitting expressions between `--raw` and non-`--raw` cells, and also leaves us open to adding `$(..)` interpolation later modulo finding a clean way to match the balanced parentheses. Thoughts?

tkf:
> The limitation is that `py"..."` can only grab stuff in the global scope.

Ah, yes, that's a good point.

---

Another idea: How about using `$"python code"`?  It seems that Julia can parse it as-is:

```
julia> Meta.parse(raw"""
       x = $"x if x is None else 0"
       """)
:(x = $(Expr(:$, "x if x is None else 0")))
```

So, we don't have to do any PyJulia-specific parsing and we can just look at the AST.

Julia parser can also do `$$` which would be useful for code-interpolation (as in`py"..."`):

```
julia> Meta.parse(raw"""
       x = $$"x if x is None else 0"
       """)
:(x = $(Expr(:$, :($(Expr(:$, "x if x is None else 0"))))))
```

tkf:
The fact that it is parsable means that someone could be using it in their macors.  So, `--raw` in your suggestion still makes sense with my `$"python code"` suggestion.

marius311:
I wanted to get feedback on the version I just force-pushed to this PR before finishing it up. What do you think of this approach? 

In this version, I got rid of `{}`, but otherwise the docs now essentially mirror PyCall:

> Any `$var` or `$(var)` expressions that appear in the Julia code (except in comments or string literals) are evaluated in Python and passed to Julia. Use `$$var` to insert a literal `$var` in Julia code.

Some examples:
```
In [1]: %load_ext julia.magic                                                                                                       
Initializing Julia interpreter. This may take some time...

In [2]: foo = [[1,2],[3,4]]                                                                                                         
Out[3]: 
array([[2, 3],
       [4, 5]], dtype=int64)

In [3]: %julia $foo .+ 1                                                                                                            

In [4]: %julia $(len(foo))
Out[4]: 2

In [5]: %julia bar=1; "$bar"   # not interpolated from Python since its in a string
Out[5]: '1'

In [6]: %julia bar=3; :($$bar)    # not interpolated from Python either
Out[6]: 3
```

The main limitation is that since I'm using Julia to parse everything, including the interpolated Python code, some small number of valid Python expressions are non-valid Julia expressions and hence don't work, e.g. `%julia $(2**3)`. If we really wanted this I could probably do something based on `pyparsing` but it's going to get way more complex, so I tend to like the simplicity of the current version. Currently also Python code which raises exceptions or calls a generator seems to crash, I can look into that. 

In any case, let me know if you have any major problems with this approach.

marius311:
Also, since it may not totally be clear what's going on, its all in this macro which parses the Julia/Python code and rewrites things so we can just call it from Python with a globals/locals dicts, e.g.:

```
julia> @macroexpand1 _PyJuliaHelper.@prepare_for_pyjulia_call begin
           $(x+1)
       end
quote
    #= REPL[1]:74 =#
    (PyCall.pyfunction)(((__globals, __locals)->begin
                #= REPL[1]:74 =#
                begin
                    #= REPL[2]:2 =#
                    (convert)(PyCall.PyAny, (PyCall.pyeval_)("x + 1", __globals, __locals))
                end
            end), PyCall.PyObject, PyCall.PyObject)
end
```

tkf:
I like the idea of using Julia's macro system.

> The main limitation is that since I'm using Julia to parse everything, including the interpolated Python code, some small number of valid Python expressions are non-valid Julia expressions and hence don't work, e.g. `%julia $(2**3)`.

Unfortunately, Julia also normalizes some valid Python code to invalid ones, for example:

```
julia> :([x for x in xs])
:([x for x = xs])
```

This is why I suggested to use `$"...."` as we can put anything in string.  (We can also re-use `py"...."` syntax instead.)

marius311:
> Unfortunately, Julia also normalizes some valid Python code to invalid ones

tkf:
I don't think it's a good idea to have multiple syntaxes to do one thing.  If `$(...)` cannot be used for all purpose, we should not have it.

Regarding `$"..."` vs `py"..."`, I have a mixed feeling.  `py"..."` sounds good because we already have it in PyCall.  However, the semantics of it is a bit different.  In PyCall, Python code is always evaluated at global scope while in PyJulia it _may_ be in the local scope if it is used like

```python
def f(x):
    y = %julia py"x + 1" + 1
    return y
```

But if we treat it as an extension of PyCall's semantics, it probably is OK?

marius311:
> I don't think it's a good idea to have multiple syntaxes to do one thing. If $(...) cannot be used for all purpose, we should not have it.

Yea, I can understand this, but `$(...)` just seems like such the natural way to do it that I wonder whether its not worth keeping anyway. As an exercise, I tried to explain the three options in the documentation, and it didn't seem to take many words to do it adequately, which I'd say is an indication it could be kept. 

> In PyCall, Python code is always evaluated at global scope while in PyJulia it may be in the local scope if it is used like

This is a good point to bring up. Actually, in the current version of this PR its totally consistent with PyCall and is always the global scope, i.e. 
```python
In [10]: x = "global"
    ...: def f():
    ...:     x = "local"
    ...:     ret = %julia py"x"
    ...:     return ret
    ...: f()
Out[10]: 'global'
```

I'd say lets keep it this way, and maybe at a later time we could introduce a `py_str` option to use the local scope, i.e. `py"x"l` which would return "local" in the above example. Presumably this could be simultaneously introduced in PyCall as well.

marius311:
Next up is tests and figuring out why Python exception crash the whole thing. Hopefully I didn't go down some doomed path with this `pyeval` thing...

marius311:
I think I'll likely need some help here @stevengj @tkf, these are internals that are beyond me. Basically any python code which gives an error crashes the interpreter (MWE below). I get different stack traces python-jl vs. natively. 

Hopefully it wasn't a bad idea to have Julia call Python back from inside of `%julia`...

<details/>

Dynamically linked Python:
```
In [1]: %load_ext julia.magic                                                                                                       
Initializing Julia interpreter. This may take some time...

In [2]: %julia py"x"                                                                                                                
fatal: error thrown and no exception handler available.
#<null>
rec_backtrace at /home/marius/src/julia/src/stackwalk.c:94
record_backtrace at /home/marius/src/julia/src/task.c:217 [inlined]
jl_throw at /home/marius/src/julia/src/task.c:417
jl_type_error_rt at /home/marius/src/julia/src/rtutils.c:118
jl_type_error at /home/marius/src/julia/src/rtutils.c:125
jl_f_setfield at /home/marius/src/julia/src/builtins.c:693
pyraise at /home/marius/.julia/packages/PyCall/h14lX/src/exception.jl:183 [inlined]
_pyjlwrap_call at /home/marius/.julia/packages/PyCall/h14lX/src/callback.jl:42
unknown function (ip: 0x7f887c055c93)
jl_fptr_trampoline at /home/marius/src/julia/src/gf.c:1864
jl_apply_generic at /home/marius/src/julia/src/gf.c:2219
pyjlwrap_call at /home/marius/.julia/packages/PyCall/h14lX/src/callback.jl:49
unknown function (ip: 0x7f887c02a56e)
_PyObject_FastCallDict at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/abstract.c:2331
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4854
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
PyEval_EvalCodeEx at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4180
function_call at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/funcobject.c:604
PyObject_Call at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/abstract.c:2261
do_call_core at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:5099 [inlined]
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3397
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
fast_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4971 [inlined]
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4851
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
_PyFunction_FastCallDict at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:5063
_PyObject_FastCallDict at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/abstract.c:2310
_PyObject_Call_Prepend at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/abstract.c:2373
PyObject_Call at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/abstract.c:2261
do_call_core at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:5099 [inlined]
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3397
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
fast_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4971 [inlined]
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4851
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
PyEval_EvalCodeEx at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4180
PyEval_EvalCode at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:731
builtin_exec_impl at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/bltinmodule.c:983 [inlined]
builtin_exec at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/clinic/bltinmodule.c.h:283
_PyCFunction_FastCallDict at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/methodobject.c:234
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4830
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
gen_send_ex at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/genobject.c:189 [inlined]
_PyGen_Send at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/genobject.c:308
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:2081
gen_send_ex at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/genobject.c:189 [inlined]
_PyGen_Send at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/genobject.c:308
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:2081
gen_send_ex at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/genobject.c:189 [inlined]
_PyGen_Send at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/genobject.c:308
_PyCFunction_FastCallDict at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/methodobject.c:209
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4830
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyFunction_FastCall at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4912 [inlined]
fast_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4947 [inlined]
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4851
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyFunction_FastCall at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4912 [inlined]
fast_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4947 [inlined]
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4851
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
fast_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4971 [inlined]
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4851
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3344
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
fast_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4971 [inlined]
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4851
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
fast_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4971 [inlined]
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4851
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyFunction_FastCall at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4912 [inlined]
fast_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4947 [inlined]
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4851
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
_PyFunction_FastCallDict at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:5063
_PyObject_FastCallDict at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/abstract.c:2310
_PyObject_Call_Prepend at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/abstract.c:2373
PyObject_Call at /tmp/python-build.20190403234951.21870/Python-3.6.6/Objects/abstract.c:2261
do_call_core at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:5099 [inlined]
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3397
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
fast_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4971 [inlined]
call_function at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4851
_PyEval_EvalFrameDefault at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:3328
_PyEval_EvalCodeWithName at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4159
PyEval_EvalCodeEx at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:4180
PyEval_EvalCode at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/ceval.c:731
run_mod at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/pythonrun.c:1025 [inlined]
PyRun_FileExFlags at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/pythonrun.c:978
PyRun_SimpleFileExFlags at /tmp/python-build.20190403234951.21870/Python-3.6.6/Python/pythonrun.c:420
run_file at /tmp/python-build.20190403234951.21870/Python-3.6.6/Modules/main.c:340 [inlined]
Py_Main at /tmp/python-build.20190403234951.21870/Python-3.6.6/Modules/main.c:810
main at python3 (unknown line)
__libc_start_main at /build/glibc-OTsEL5/glibc-2.27/csu/../csu/libc-start.c:310
_start at python3 (unknown line)
```

and python-jl:

```
In [2]: %julia py"x"                                                                                                                
ERROR: TypeError: in setfield!, expected Ptr{PyCall.PyObject_struct}, got Ptr{Nothing}
Stacktrace:
 [1] pyraise at /home/marius/.julia/packages/PyCall/h14lX/src/exception.jl:183 [inlined]
 [2] _pyjlwrap_call(::PyCall.FuncWrapper{Tuple{PyObject,PyObject},getfield(Main, Symbol("##3#4"))}, ::Ptr{PyCall.PyObject_struct}, ::Ptr{PyCall.PyObject_struct}) at /home/marius/.julia/packages/PyCall/h14lX/src/callback.jl:42
 [3] pyjlwrap_call(::Ptr{PyCall.PyObject_struct}, ::Ptr{PyCall.PyObject_struct}, ::Ptr{PyCall.PyObject_struct}) at /home/marius/.julia/packages/PyCall/h14lX/src/callback.jl:49
 [4] macro expansion at /home/marius/.julia/packages/PyCall/h14lX/src/exception.jl:81 [inlined]
 [5] __pycall!(::PyObject, ::Ptr{PyCall.PyObject_struct}, ::PyObject, ::Ptr{Nothing}) at /home/marius/.julia/packages/PyCall/h14lX/src/pyfncall.jl:44
 [6] _pycall!(::PyObject, ::PyObject, ::Tuple{Array{String,1}}, ::Int64, ::Ptr{Nothing}) at /home/marius/.julia/packages/PyCall/h14lX/src/pyfncall.jl:29
 [7] #call#111 at /home/marius/.julia/packages/PyCall/h14lX/src/pyfncall.jl:11 [inlined]
 [8] (::PyObject)(::Array{String,1}) at /home/marius/.julia/packages/PyCall/h14lX/src/pyfncall.jl:89
 [9] top-level scope at none:0
Exception ignored in: <module 'threading' from '/home/marius/.pyenv/versions/3.6.6/lib/python3.6/threading.py'>
  File "/home/marius/.pyenv/versions/3.6.6/lib/python3.6/threading.py", line 1289, in _shutdown
    assert tlock.locked()
SystemError: <built-in method locked of _thread.lock object at 0x7f13a4ded170> returned a result with an error set
```

tkf:
I'm strongly against to have multiple options because

> [There should be one-- and preferably only one --obvious way to do it.](https://www.python.org/dev/peps/pep-0020/)

Using Julia's `string(::Expr)` is a good experiment but it seems to be too much of a hack to be a part of a package.  I prefer to not release half-baked features especially if there is no one working on it to complete it (e.g., writing a special parser to do this correctly).

> totally consistent with PyCall and is always the global scope, i.e.
> 
> ```python
> In [10]: x = "global"
>     ...: def f():
>     ...:     x = "local"
>     ...:     ret = %julia py"x"
>     ...:     return ret
>     ...: f()
> Out[10]: 'global'
> ```

Thinking about this more, I think this is not a useful behavior and is potentially a big pitfall for Python users.  I started thinking `py"..."` is a good idea because we can fix this pitfall and then introduce a useful variable interpolation at the same time.  You can do `PyCall.py"..."` if you need PyCall behavior.

tkf:
Re: the error with `%julia py"x"`; I can't reproduce it. Can you open a new issue with all the relevant version info (julia, python, pyjulia, PyCall)?

marius311:
> I'm strongly against to have multiple options because

That's fine, I can remove it. 

> Thinking about this more, I think this is not a useful behavior and is potentially a big pitfall for Python users. I started thinking py"..." is a good idea because we can fix this pitfall and then introduce a useful variable interpolation at the same time. You can do PyCall.py"..." if you need PyCall behavior.

I didn't understand this, sorry. 

> Re: the error with %julia py"x"; I can't reproduce it. Can you open a new issue with all the relevant version info (julia, python, pyjulia, PyCall)?

This is with this PR, so it seems here is the right place to discuss it? Otherwise I have julia 1.1.0, python 3.6.6, and PyCall v1.91.1.

tkf:
Which is the part you don't understand?  I'm just saying that we can "hide" the current misbehavior of `py"..."` (not aware of Python's local scope) by implementing the variable interpolation as you did for `$(...)`.

> This is with this PR

Ah, sorry.  I thought it happened without this PR.

tkf:
I can reproduce it.

tkf:
https://github.com/JuliaPy/PyCall.jl/pull/675 fixed it for me

src/julia/magic.py
```diff
@@ -22,7 +22,7 @@
 import sys
 import warnings
 
-from IPython.core.magic import Magics, magics_class, line_cell_magic
+from IPython.core.magic import Magics, magics_class, line_cell_magic, no_var_expand
```
tkf: Let's support older IPython with

```python
try:
    from IPython.core.magic import no_var_expand
except ImportError:
    def no_var_expand(f):
        return f
```

(put it after `from .tools import redirect_output_streams`)

tkf:
(JuliaPy/PyCall.jl#675 accidentally closed this issue)

marius311:
> Which is the part you don't understand? I'm just saying that we can "hide" the current misbehavior of py"..." (not aware of Python's local scope) by implementing the variable interpolation as you did for $(...).

Ah sorry I had thought PyCall also "misbehaved" like this but I realize now it works correctly. I fixed it here as well, see the updated doc which now returns "local".

Trying to fix test failures now...

tkf:
You can use `tox -- --julia-compiled-modules=no` etc. to run tests locally (even if you don't have dynamically linked Python) now if you merge `master` to your branch.  See:
https://pyjulia.readthedocs.io/en/latest/testing.html
https://pyjulia.readthedocs.io/en/latest/pytest.html#options

marius311:
I think its ready from my end.

src/julia/pyjulia_helper.jl
```diff
@@ -44,6 +48,53 @@ if VERSION >= v"0.7-"
...
+    # into the new expression
+    function stoppable_walk(f, x)
+        (fx, recurse) = f(x)
+        walk(fx, (recurse ? (x -> stoppable_walk(f,x)) : identity), identity)
+    end
+    
+    locals = gensym("locals")
+    globals = gensym("globals")
+    
+    function make_pyeval(expr, options...)
```
tkf:
```suggestion
    function make_pyeval(expr::Union{String,Symbol}, options...)
```

seems to be safer as we are doing `code = string(expr)`.

docs/source/usage.md
```diff
@@ -67,23 +67,57 @@ You can then use, e.g.,
...
+```
+
+Python interpolation is not performed inside of strings (instead this is treated as regular Julia string interpolation), and can also be overridden elsewhere by using `$$...` to insert a literal `$...`:
+
+```julia
+In [6]: %julia foo=3; "$foo"
+Out[6]: '3'
+
+In [7]: %julia bar=3; :($$bar)
+Out[7]: 3
```
tkf: I think the behavior comparable to PyCall is

```
In [6]: bar = "3"
   ...: %julia $$bar ^ 4
Out[6]: "3333"

In [7]: %julia $$foo ^ 4
Out[7]: 81
```

See:

```julia
julia> bar = "3"
       py"$bar * 4"
"3333"

julia> bar = "3"
       py"$$bar * 4"
12
```

test/test_magic.py
```diff
@@ -1,11 +1,36 @@
...
+
 import pytest
+from IPython import get_ipython as _get_ipython
+from IPython.testing.globalipapp import start_ipython as _start_ipython
+
+from julia import magic
 
 
+def get_ipython():
+    return _start_ipython() or _get_ipython()
```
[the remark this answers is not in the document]
marius311: `IPython.testing.globalipapp.get_ipython` returns `None` after the first time its been called (not sure if bug?) and that was happening with this fixture thing, leading to `juliamagic.shell.user_ns` (the global dict in the variable interpolation) being an error. Not sure if there's a better way to get a shell in testing?

test/test_magic.py
```diff
@@ -1,11 +1,36 @@
...
+        if cell.startswith("%%"):
+            return self.shell.run_cell_magic("julia","",cell.replace("%%julia","").strip())
+        else:
+            exec_result = self.shell.run_cell(cell)
+            if exec_result.error_in_exec:
+                raise exec_result.error_in_exec
+            else:
+                return exec_result.result
+                
+    julia_magics.run_cell = run_cell.__get__(julia_magics, julia_magics.__class__)
```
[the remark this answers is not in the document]
marius311: Yea I like your first suggestion, will go with that.

docs/source/usage.md
marius311: Did you have an extra `$` in your In[6]? I think that first case is totally consistent between Python/Julia currently: 

```
julia> bar = "3"
       py"$bar * 4"
"3333"

In [15]: bar = "3" 
    ...: %julia $bar ^ 4                                                                                                            
Out[15]: '3333'
```

The other case is indeed different because PyCall chooses to have `$$foo` mean interpolate foo as string into the Python expression, whereas I chose to have it mean a literal `$foo`. I think it makes some sense these can be different because `$foo` is valid Julia but not Python syntax. In any case, I guess the usecase of `py"$$foo"` which is to use Julia to build Python expressions seems really unlikely to be used in the other direction, whereas wanting a literal `$foo` seems much more common, so I'd be apt to keep it as is. Thoughts?

tkf: > Did you have an extra `$` in your In[6]?

Oops. Good catch. I meant to write `%julia $bar ^ 4`. Fixed now.

tkf: > wanting a literal `$foo`

it can be done with

```
foo = "$foo"
%julia $$foo
```

though I guess this is a bit strange.

The literal `$foo` in Julia is valid only in macro and quote contexts. I think we need to stop replacing `$` and `$$` inside those contexts.  We can still use `py"..."` everywhere so I think it's OK.

test/test_magic.py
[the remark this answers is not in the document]
tkf: This is how `IPython.testing.globalipapp.get_ipython` is defined

https://github.com/ipython/ipython/blob/4f901de17aff2132365fc3741fc5b4b8a9a98dc6/IPython/testing/globalipapp.py#L48-L50
https://github.com/ipython/ipython/blob/4f901de17aff2132365fc3741fc5b4b8a9a98dc6/IPython/testing/globalipapp.py#L119-L124

So, I think we should be using like this

```python
from IPython.testing import globalipapp

# then
magic.JuliaMagics(shell=globalipapp.get_ipython())
```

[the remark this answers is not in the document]
marius311: That works nicely, thanks.

docs/source/usage.md
marius311: I think I'll have to think about this one over the weekend. My issue with stopping replacing `$` and `$$` in macros and quotes is two fold:

1) not all macros expand `$foo`, e.g. `@eval $foo` does but `@time $foo` does not, so I don't think it makes sense to stop expanding in all macros, and it may be hard/impossible to detect which ones do expand
2) in quotes, what about `:($($foo))` where logically `$foo` should be interpolated? you could do this, but then you'd have to track the "quote depth" which would be more complex.

For these reasons I thought the current rule was both simpler to remember and to code up. Then again, I guess the depth thing *is* currently handled in string interpolation (because Julia did for me for free). I.e. `%julia "$($foo)"` interpolates `foo` from Python. So maybe it would make most sense to work in expressions too so that `%julia :($($foo))` worked and interpolated `$foo`...

tkf: > For these reasons I thought the current rule was both simpler to remember and to code up.

I argue that a _valid_ Julia code snippet has to be evaluated correctly in `%%julia` magic [1].  Assigning `$$` a different meaning in `%%julia` violates this.  I think it is the best to make `%%julia` a _superset_ of the Julia language; i.e., additionally allow `$foo` and `$$bar` in the "toplevel" contexts which is not the case in standard Julia.  Simple to use does not mean simple to implement.  I'm all for having a few AST-manipulating functions to help users.

> 1. not all macros expand `$foo`, e.g. `@eval $foo` does but `@time $foo` does not

I'd say we should play on the safe side.  It's impossible to list all cases where `@somemacro $foo` is valid or not.  So it's safer to stop touching the AST as soon as we see a macro call.

> 2\. in quotes, what about `:($($foo))` where logically `$foo` should be interpolated?

Yes, this is a good point.  We better do this.  But I think leaving it to throw an error at this point is OK, too.  Users' code will be forward compatible to new PyJulia supporting it.

Maybe the easiest thing here is to drop `$foo` from _this_ PR and do it afterwards?  We can still use `py"..."` for all contexts, right?  It is always much easier to change the syntax that was an error to do something useful than change the behavior that breaks users' code.

[1] Side note: Local `py"..."` does not violate this because it would not be valid in vanilla Julia + PyCall.  But maybe this point is a bit subtle.

marius311:
After thinking about it I think you're right about
> a valid Julia code snippet has to be evaluated correctly in %%julia magic 


This means interpolation from Python via `$` should only happen outside of strings and quotes, since inside of them (and only inside), `$foo` already has a valid Julia meaning. Similarly, in pure Julia, `:(py"...")` would *not* call PyCall rather it would just create that expression object, so inside of `%julia` magic that shouldn't interpolate either. So the rule now is basically:

> Interpolation from Python is only done outside of strings and quote blocks

This leads to things like
```
In [6]: foo="Python"
        %julia foo="Julia"
        %julia "this is $foo", "this is $($foo)"
Out[6]: ('this is Julia', 'this is Python')
```

which I think is entirely natural. I added the logic to track the "quote depth" which you need to do this to the PR, it was really simple. I also turned off interpolation inside of macros entirely, I think you're right that there you just have to err on the safe side. And I removed `$$` entirely, that seems unnecessary now; it could potentially be given PyCall-like behavior later if we want to. Ready for any more comments you may have.

tkf:
This PR looks great now.  Thanks a lot for implementing it!

> So the rule now is basically:
> 
> > Interpolation from Python is only done outside of strings and quote blocks

This sounds good to me.  Thanks for noticing the string case (which I forgot).  The example looks good.

---

Before merging, a bit of nitpicking of the wording...

Doesn't the word "interpolation" suggest that the following code prints the same number five times?

```julia
for _ in 1:5
    println(py"numpy.random.rand()")
end
```

Also, if we interpret "interpolation" as "pre-evaluation object injection" (as in `@benchmark` and `PyCall.py"..."`), the semantics `$x` of is a bit different from usual `:($x)`:

```julia
In [2]: x = 1

In [3]: %%julia
        println($x)
        py"""
        x = 2
        """
        println($x)
```

which print 1 and 2 in this PR but with "pre-evaluation object injection" semantics it would be 1 twice.

This is fine by me (although I'm somewhat nervous a little bit) as we can do whatever we want do to define the semantics of depth-1 `$` and it's more powerful than "pre-evaluation object injection."

Maybe we can just say

> You can "interpolate" Python **code** into Julia code

instead of _Python **objects**_?  Or is there a better phrase for it?

marius311:
Fair point, how about:

> You can call Python code from inside of `%julia` blocks via `$var` for accessing single variables or `py"..."` for more complex expressions:

marius311:
FWIW I find both Julia and PyCall behavior totally weird compared to what we have here. In Julia it looks like its always the final value of the interpolated thing (if it was going to be a fixed thing wouldn't the initial value make more sense?), e.g:

```
julia> x=1
1

julia> "$x $(x=2) $(x=3) $x $(x=4)"
"4 2 3 4 4"

julia> x
4
```

and PyCall is maybe trying to be that but I think there's got to be a bug?

```
julia> x=1
1

julia> py"""
       print($x)
       $(x=2)
       print($x)
       """
2
2

julia> x
2

julia> x=1
1

julia> py"""
       print($x)
       $(x=2)
       print($x)
       """
1
2
```

tkf:
> Fair point, how about:
> 
> > You can call Python code from inside of `%julia` blocks via `$var` for accessing single variables or `py"..."` for more complex expressions:

Ah, this sounds more explicit and easy to understand.  It'd be great if you can tweak the docs to avoid using the word "interpolation."

---

BTW the examples you brought up is interesting.  I didn't know that's what Julia do.  FYI this is what Julia evaluates:

```
julia> Meta.@lower "$x $(x=2) $(x=3) $x $(x=4)"
:($(Expr(:thunk, CodeInfo(
    @ none within `top-level scope'
1 ─      x = 2
│        x = 3
│        x = 4
│   %4 = Base.string(x, " ", 2, " ", 3, " ", x, " ", 4)
└──      return %4
))))
```

But I have no idea why this happens.  Maybe report a bug to Julia?

OTOH, PyCall seems to not track the order (maybe because `Dict` forgets the order?)

```
julia> Meta.@lower py"""
       print($x)
       $(x=2)
       print($x)
       """
:($(Expr(:thunk, CodeInfo(
    @ /home/takafumi/.julia/packages/PyCall/h14lX/src/pyeval.jl:230 within `top-level scope'
1 ─ %1  = PyCall.pynamespace(Main)
│         #27#m = %1
│   @ /home/takafumi/.julia/packages/PyCall/h14lX/src/pyeval.jl:231 within `top-level scope'
│         x = 2
│   %4  = PyCall.PyObject(2)
│         Base.setindex!(#27#m, %4, "__julia_localvar_0_2")
│   %6  = PyCall.PyObject(x)
│         Base.setindex!(#27#m, %6, "__julia_localvar_0_3")
│   %8  = PyCall.PyObject(x)
│         Base.setindex!(#27#m, %8, "__julia_localvar_0_1")
│   @ /home/takafumi/.julia/packages/PyCall/h14lX/src/pyeval.jl:232 within `top-level scope'
│   %10 = Base.getproperty(Base, :string)
│   %11 = (%10)("print(__julia_localvar_0_1)\n__julia_localvar_0_2\nprint(__julia_localvar_0_3)\n")
│   %12 = PyCall.pyeval_(%11, #27#m, #27#m, 257, "/home/takafumi/.julia/packages/PyCall/h14lX/src/pyeval.jl")
│   %13 = (PyAny)(%12)
│         #28#ret = %13
│   @ /home/takafumi/.julia/packages/PyCall/h14lX/src/pyeval.jl:234 within `top-level scope'
└──       return #28#ret
))))
```

stevengj:
> In Julia it looks like its always the final value of the interpolated thing (if it was going to be a fixed thing wouldn't the initial value make more sense?), ... and PyCall is maybe trying to be that but I think there's got to be a bug?

They are a little bit different.  Julia `$foo` string interpolation interpolates values as strings.  PyCall `$foo` interpolation in `py"..."` interpolates Julia *objects*.   In that sense, it is more like `Expr` interpolation in Julia:
```jl
julia> x = 1
1

julia> :($x + $(x=2) + $x)
:(2 + 2 + 2)
```
That being said, `py"..."` could probably be modified to call `PyObject(arg)` on each argument before evaluating the next argument if we really wanted that.  It wouldn't change the behavior for mutable objects like arrays, however — changes to an array from one argument would still be "visible" to all other references to the same array.

tkf:
@marius311 Thanks for your awesome work!

tkf:
@stevengj Would it be fixed by just using `locals = Pair{Union{String,Int},Any}[]` instead of `locals = Dict{Union{String,Int},Any}()` in `interpolate_pycode`?

marius311:
Thanks for all the comments, this was way better because of them!

marius311:
Super minor but any idea what happened to the syntax highlighting on cells 6 & 7 in https://pyjulia.readthedocs.io/en/latest/usage.html#ipython-magic ? It looks right when I build the docs locally. Note those are marked language Julia bc the `$` screwed up the highlighting if it was marked python (at least, that's locally what happened)

tkf:
Hmm... Good point. I have no idea. Maybe the package versions are different?  Though looking at some of the packages listed in https://readthedocs.org/projects/pyjulia/builds/8884519/ they seem to use fairly new ones.  We can in principle freeze the version of all the packages by putting them in `docs/requirements.txt`.  But I guess it's better to switch to IPython Sphinx Directive https://ipython.readthedocs.io/en/stable/sphinxext.html than fixing this.  We can use it as doctest too.

PR #265, opened 2019-04-03
Do not swallow exceptions in %julia magic

tkf:
Errors should never pass silently.
Unless explicitly silenced.

(see #266 for full context)


tkf:
#262 did it

Issue #268, opened 2019-04-05
Auto-install python-jl Jupyter kernel? 


marius311:
Would you accept a PR that makes it so that upon installing PyJulia, a python-jl-based Jupyter kernel is installed (if Jupyter is present)? I can write this.

tkf:
I'm against for _auto_-install because

* Even ipykernel does not install the kernel https://ipython.readthedocs.io/en/stable/install/index.html
* It is very common to have multiple Python virtual environments.  I think it would be impossible to sanely handle situation where you install one kernel in environment A and then after in environment B.
* I want to get rid of `python-jl` at some point (not very soon, though). I think https://github.com/JuliaPy/pyjulia/pull/256 and https://github.com/JuliaPy/PyCall.jl/issues/612 are the better way.

Having said that, I think it is a good idea to have something equivalent to `python -m ipykernel install` (say `python -m julia.ipykernel install --python-jl`).  We can also solve #223 with this.

marius311:
I'm still interested in the kernel installer, that's fine that its not auto-installed. Should I work on that (I think it's not much work) or do you think it's not worth it given how close you may be on getting rid of `python-jl` entirely?

tkf:
I think it's still better to have `python -m julia.ipykernel install` command due to #223.  Adding `--python-jl` flag for this command to choose `python-jl` as a "backend" sounds good to me (unlike `python-jl -m IPython`, there is no obvious command to do this ATM).

marius311:
> I think it's still better to have python -m julia.ipykernel install

Sounds good, that's what I'll aim for.

PR #256, opened 2019-03-25
Generate system image with PyCall in "PyJulia-mode"

tkf:
closes #217

This requires:

* [x] PyCall release with https://github.com/JuliaPy/PyCall.jl/pull/651
* ideally https://github.com/JuliaLang/julia/pull/29914 and https://github.com/JuliaLang/julia/pull/30065; but it turned out we can gorilla patch them while compiling the system image https://github.com/JuliaPy/pyjulia/pull/256/commits/41df92e6590a2ce12a07dfd4a973e67ff43646ec
* [x] Remove `[DO NOT MERGE]` commits

PR #269, opened 2019-04-05
add ipython-jl

marius311:
Seemed a more seamless experience for Pythonistas.

### DIFF
--- a/docs/source/troubleshooting.md
+++ b/docs/source/troubleshooting.md
@@ -39,7 +39,12 @@ basic use-cases such as:
 ```console
 $ python-jl your_script.py
 $ python-jl -c 'from julia.Base import banner; banner()'
-$ python-jl -m IPython
+```
+
+For IPython, use:
+
+```console
+$ ipython-jl
 ```
 
 See `python-jl --help` for more information.
--- a/setup.py
+++ b/setup.py
@@ -70,6 +70,7 @@ setup(name='julia',
       entry_points={
           "console_scripts": [
               "python-jl = julia.python_jl:main",
+              "ipython-jl = julia.python_jl:ipython_main",
           ],
           "pytest11": [
               "julia = julia.pytestplugin",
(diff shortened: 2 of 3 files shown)


tkf:
I appreciate your interest for improving PyJulia.  But I'd like to minimize the interface related to `python-jl` as there would be better ways to support PyJulia in wider range of Python installations (see also https://github.com/JuliaPy/pyjulia/issues/268#issuecomment-480128107).  I hope you are not discouraged because I'm closing this.

tkf:
BTW, checkout https://github.com/tkf/IPython.jl if you want a "better" version `python-jl -m IPython`.  I think it's strictly better than IPython for using PyJulia because you have access to both IPython and Julia REPL (plus some event loop coordination).

marius311:
Thanks, appreciate the thoughtful response. You brought up good points about the Jupyter kernel, will reply in that issue, but here I really think you should reconsider. I totally agree `python-jl` should go away as soon as it can, but currently it does have to exist so why not just make things slightly more userfriendly where we can? Its not as if this is some API that will be hard to deprecate later, its just what people type to launch an interpreter.

tkf:
I'm very close to implementing an "ultimate solution" #256.  Basically, the blockers are dropping Julia 0.6 support (which in turn is blocked by PyJulia 0.3 release) and new PyCall release.  Once this is implemented, the information related to `ipython-jl` would be harmful as it would misguide users.  The reason why I said I'll not get rid of `python-jl` very soon was not because `python-jl` will still be a valid solution but because to keep backward compatibility for a while.

Due to the limited resource in development, we need to focus on good solutions.  `python-jl` is not a good solution because it is impossible to make it right.  There is no guarantee that the Python environment that `python-jl` is installed in and the Python environment active inside `python-jl` session are the same.  This would be extremely confusing for people who are not familiar with how Python imports work.

marius311:
Ok fair enough, I didn't realize the ultimate solution was so close. Looking forward to that!

PR #286, opened 2019-04-14
Bump to 0.5.0.dev

tkf:


docs/source/conf.py
```diff
@@ -24,9 +24,9 @@
 author = 'The Julia and IPython development teams'
 
 # The short X.Y version
-version = '0.4.0'
+version = '0.5.0'
 # The full version, including alpha/beta/rc tags
-release = '0.4.0'
+release = '0.5.0.dev'
```
stevengj: The usual semver style is `0.5.0-dev`

stevengj: Oh, I see — https://www.python.org/dev/peps/pep-0440/#version-scheme does indeed require `.dev`

PR #295, opened 2019-04-22
Automatically deploy package to test.pypi.org

tkf:
This patch adds two stages to release the package at test.pypi.org and then test the deployed package.  These two stages are triggered when there is a push to the special branch `release/test`.  The package uploaded to test.pypi.org can then be released on the standard pypi.org server by locally running `pip download` and `twine` (via `make release`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,11 +41,48 @@ matrix:
           - JULIA_VERSION=0.6.4
           - CROSS_VERSION=1
         os: osx
+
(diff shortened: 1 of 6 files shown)


tkf:
Example:
https://travis-ci.org/JuliaPy/pyjulia/jobs/523283894#L804-L808
https://test.pypi.org/project/julia/0.4.1.dev0/

PR #299, opened 2019-04-26
Refactoring: Pull JuliaInfo/LibJulia out of julia.core

tkf:
(to be rebased after #298)

### DIFF
--- a/src/julia/core.py
+++ b/src/julia/core.py
@@ -613,6 +613,8 @@ class Julia(object):
         ctypes.pythonapi.Py_IncRef(ctypes.py_object(pyobj))
         return pyobj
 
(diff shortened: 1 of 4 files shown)


docs/source/installation.rst
```diff
@@ -115,7 +115,6 @@ Alternatively, you can use Julia's builtin package manager.
 
 .. code-block:: jlcon
 
-   julia> using Pkg  # for julia ≥ 0.7
```
PallHaraldsson: using Pkg
is still needed, or ]add PyCall

tkf: Indeed. Thanks a lot!

PR #312, opened 2019-07-30
Added user site bin for julia-py

krystophny:
Since I'm using `setup.py develop --user` in my setup, `julia-py` binary is installed in `~/.local/bin`. Rather than hardcoding the path to be at the same place as the `python` binary, one should look also there. This patch adds this behavior. Do you think one could also use `site.PREFIXES` to replace the default binary path? In that case I will create another patch.

### DIFF
--- a/src/julia/tools.py
+++ b/src/julia/tools.py
@@ -2,9 +2,11 @@ from __future__ import absolute_import, print_function
 
 import glob
 import os
+import platform
 import re
 import site
 import subprocess
+import sysconfig
 import sys
 
 from .core import JuliaNotFound, which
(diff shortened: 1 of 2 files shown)


src/julia/tools.py
```diff
@@ -139,8 +140,11 @@ def julia_py_executable(executable=sys.executable):
     """
     Path to ``julia-py`` executable installed for this Python executable.
     """
-    stempath = os.path.join(os.path.dirname(executable), "julia-py")
-    candidates = {os.path.basename(p): p for p in glob.glob(stempath + "*")}
+    basedirs = {os.path.dirname(executable),
+                os.path.join(site.USER_BASE, "bin")}
```
tkf: Why use set? I think we should use a list here in the order of increasing preference (so that the values of `candidates` are the one in the most preferred location).  Does `basedirs = [os.path.dirname(executable) + os.path.join(site.USER_BASE, "bin")] + site.PREFIXES` make sense?

tkf: Actually, `os.path.join(site.USER_BASE, "bin")` may not work in Windows?  Is there a pre-defined path we can use?

krystophny: Thanks for the hint! I've changed the logics according to your remarks. On Windows there seem to be more difficulties anyway (starting with "libjulia" library extension), but this part looks like it should work if the rest does too.

krystophny:
Closing, since 5 years without anyone checking further.

MilesCranmer:
@krystophny please check out [PythonCall.jl](https://github.com/JuliaPy/PythonCall.jl) which is actively developed

PR #320, opened 2019-09-11
honor trailing semicolon in Julia magic to suppress return value

marius311:
Hopefully self explanatory title :)

### DIFF
--- a/src/julia/magic.py
+++ b/src/julia/magic.py
@@ -115,13 +115,15 @@ class JuliaMagics(Magics):
         while caller_frame.f_globals.get('__name__').startswith("IPython"):
             caller_frame = caller_frame.f_back
         
-        ret_value = self._julia.eval("""
-        _PyJuliaHelper.@prepare_for_pyjulia_call begin %s end
-        """%src)(self.shell.user_ns, caller_frame.f_locals)
+        return_value = "nothing" if src.strip().endswith(";") else ""
         
-        if not src.strip().endswith(";"):
-            return ret_value
-
+        return self._julia.eval("""
+        _PyJuliaHelper.@prepare_for_pyjulia_call begin 
+            begin %s end
+            %s
+        end
+        """%(src,return_value))(self.shell.user_ns, caller_frame.f_locals)
+        
         # fmt: on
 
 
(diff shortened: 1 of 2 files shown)


marius311:
Actually, there's something weird going on where it still tries to pyconvert the answer even if there's a semicolon, looking into it...

tkf:
Hi, thanks for the fix. Can you set up https://pre-commit.com/ so that it does not fail on style check like this? https://travis-ci.org/JuliaPy/pyjulia/jobs/583925058

tkf:
The title says WIP. Do you want to do something else here?

KristofferC:
Maybe I am misunderstanding but `;` shouldn't, in general, do anything to the return value. It just stops something being *printed*.

marius311:
I think you're right. I did this with cell magic in mind, but for line magic its more subtle. To be most consistent, you want `x = %julia 1` to set `x` to 1 and print 1, but `x = %julia 1;` to not print anything, but still set `x` to 1. Do you guys know if that's actually possible given the magic API? It doesn't seem immediately obvious, but I can take a look this week.

tkf:
@KristofferC Thanks for pointing it out.  I didn't notice this subtlety.

Note that IPython treats `;` to mean "not returning this value":

```
In [1]: "hello"
Out[1]: 'hello'

In [2]: "world";

In [3]: print(_)
hello

In [4]: 1 in Out
Out[4]: True

In [5]: 2 in Out
Out[5]: False
```

unlike Julia:

```
julia> "hello"
"hello"

julia> "world";

julia> println(ans)
world
```

I think the question is what `;` at the end of IPython cell magic should mean.  Does it mean IPython's `;` or Julia's `;`?  I think treating `;` as in this PR for the cell magic is most practical option because the printing is handled by IPython in this case and following IPython's convention sounds reasonable.

> To be most consistent, you want `x = %julia 1` to set `x` to 1 and print 1, but `x = %julia 1;` to not print anything, but still set `x` to 1. Do you guys know if that's actually possible given the magic API?

@marius311 I don't know how to do this. My guess is it would be a very hacky solution.  Maybe the safest option would be to not do any `;` handling in the line magic.  People can always write

```
x = %julia 1
None
```

PR #324, opened 2019-09-16
Fixed specified procedure not found windows

### LINKED ISSUE #67: WindowsError "The specified procedure could not be found" on initializing pyjulia object
Hi,

I am having an issue starting pyjulia on Windows 10. When I run 

j = julia.Julia() 

I receive the following dialog saying:

The procedure entry point _ZSt24_throw_out_of_range_fmtPKcz could not be located in the dynamic link library C:...\Julia-0.5.0\bin\LLVM-3.7.dll

This is followed by another dialog with the same message regarding libjulia.dll

I am using Version 0.5.0 (2016-09-19 18:14 UTC  |  x86_64-w64-mingw32    of Julia.

I believe I have followed your installation instructions .For installing PyCall, I set the python environment to '' and PyCall downloaded a conda installation. It may have downloaded a 32 bit version of python. 

Thanks for your help in advance.

darrencl:
This is achieved by:
* Change working dir to libjulia folder upon calling libjulia.dll

Fix #67

### DIFF
--- a/src/julia/install.jl
+++ b/src/julia/install.jl
@@ -64,7 +64,7 @@ catch err
 end
 
 
-ENV["PYTHON"] = python
+ENV["PYTHON"] = lowercase(python)
 
 # TODO: warn if some relevant JULIA_* environment variables are set
 # TODO: use PackageSpec to specify PyCall's UUID
--- a/src/julia/install.jl
+++ b/src/julia/install.jl
@@ -64,7 +64,7 @@ catch err
 end
 
 
-ENV["PYTHON"] = lowercase(python)
+ENV["PYTHON"] = python
 
 # TODO: warn if some relevant JULIA_* environment variables are set
 # TODO: use PackageSpec to specify PyCall's UUID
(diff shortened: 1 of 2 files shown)


src/julia/libjulia.py
```diff
@@ -210,7 +210,10 @@ def __init__(self, libjulia_path, bindir, sysimage):
         if sys.version_info >= (2, 7, 13) and sys.version_info < (2, 7, 14):
             libjulia_path = libjulia_path.encode("ascii")
 
+        cwd = os.getcwd()
+        os.chdir(os.path.dirname(libjulia_path))
         self.libjulia = ctypes.PyDLL(libjulia_path, ctypes.RTLD_GLOBAL)
+        os.chdir(cwd)
```
tkf: We need to do this inside `try`-`finally`

```python
        try:
            os.chdir(os.path.dirname(libjulia_path))
            self.libjulia = ctypes.PyDLL(libjulia_path, ctypes.RTLD_GLOBAL)
        finally:
            os.chdir(cwd)
```

src/julia/install.jl
```diff
@@ -64,7 +64,7 @@ catch err
 end
 
 
-ENV["PYTHON"] = python
+ENV["PYTHON"] = lowercase(python)
```
tkf: `python` should be normalized on the Python side. Also, I think using `lowercase` would break for non-Windows users.

Also, why is this necessary? Please see https://github.com/JuliaPy/pyjulia/issues/67#issuecomment-480649375

darrencl: Ah! Yep, totally forgot on that. Thanks for pointing out.

PR #353, opened 2020-01-19
Run tests using GitHub Actions

tkf:
Run tests on GitHub Actions for

* Python 3.7, 3.8
* Julia 1.3, 1.5
* Linux, macOS, Windows

(with some combinations excluded)

Let's keep AppVeyor for now since there is one test that fails on GitHub (so it's skipped for now) but not in AppVeyor.  Also keeping Travis for now since the CI is pretty fragile...

### DIFF
--- a/src/julia/tests/test_python_jl.py
+++ b/src/julia/tests/test_python_jl.py
@@ -10,6 +10,8 @@ from julia.core import which
 from julia.python_jl import parse_pyjl_args
 from julia.utils import is_apple
 
+from .utils import skip_in_github_actions_windows
+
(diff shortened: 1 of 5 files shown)


tkf:
We need https://github.com/actions/virtual-environments/issues/125 for macOS

src/julia/tests/test_python_jl.py
```diff
@@ -67,6 +69,7 @@ def test_cli_quick_pass_no_julia(args):
 
 
 @python_jl_required
+@skip_in_github_actions_windows
```
tkf, line 72: Ignore the failure in `test_cli_import` for now... https://github.com/JuliaPy/pyjulia/runs/934005170#step:8:133

Issue #355, opened 2020-01-29
**from julia import Base** crash


rdguerrerom:
Hi, 
 I am trying to use the package as you recommend in the README without success. I am using ipython.  First I installed the julia package in python:
    
    pip3 install--user julia
 
test:

    python-jl
    ERROR: PyError (PyImport_ImportModule
    
    The Python package julia.pseudo_python_cli could not be found by pyimport. Usually this means
    that you did not install julia.pseudo_python_cli in the Python version being used by PyCall.
```

PyCall is currently configured to use the Julia-specific Python distribution
installed by the Conda.jl package.  To install the julia.pseudo_python_cli module, you can
use `pyimport_conda("julia.pseudo_python_cli", PKG)`, where PKG is the Anaconda
package the contains the module julia.pseudo_python_cli, or alternatively you can use the
Conda package directly (via `using Conda` followed by `Conda.add` etcetera).

Alternatively, if you want to use a different Python distribution on your
system, such as a system-wide Python (as opposed to the Julia-specific Python),
you can re-configure PyCall with that Python.   As explained in the PyCall
documentation, set ENV["PYTHON"] to the path/name of the python executable
you want to use, run Pkg.build("PyCall"), and re-launch Julia.

) <class 'ModuleNotFoundError'>
ModuleNotFoundError("No module named 'julia'")

Stacktrace:
 [1] pyimport(::String) at /Users/ruben/.julia/packages/PyCall/ttONZ/src/PyCall.jl:544
 [2] top-level scope at none:4
```
Testing with ipython:

```
$ ipython
Python 3.7.4 (default, Aug 13 2019, 15:17:50)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.8.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import julia

In [2]: julia.install()
[ Info: Julia version info
Julia Version 1.3.1
Commit 2d5741174c (2019-12-30 21:36 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.6.0)
  uname: Darwin 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64 x86_64 i386
  CPU: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz:
              speed         user         nice          sys         idle          irq
[...]

Issue #336, opened 2019-11-26
Problems compiling Julia system image in a Docker


opus111:
As described here [Compiling Julia system image in a Docker](https://stackoverflow.com/questions/59059441/compiling-julia-system-image-in-a-docker/59060055#59060055) julia-py has problems when run as part of building a Docker.

Python is installed at /usr/bin but the PyJulia executables are installed at /usr/local/bin.  When julia-py is executed if fails with
`RuntimeError: ``julia-py`` executable is not found for Python installed at /usr/bin/python3`

PR #366, opened 2020-04-15
correct the way of finding Python's scripts folder

thautwarm:
When calling julia-py, `julia.core.julia_py_executable` is invoked, and currently works correctly for limited Unix distributions only. See

- https://github.com/python/cpython/blob/master/Lib/sysconfig.py
- https://github.com/python/cpython/blob/master/Doc/install/index.rst

I improved this, hence it shall then work for every OS supported by Python.

### DIFF
--- a/src/julia/tools.py
+++ b/src/julia/tools.py
@@ -139,7 +139,9 @@ def julia_py_executable(executable=sys.executable):
     """
     Path to ``julia-py`` executable installed for this Python executable.
     """
-    stempath = os.path.join(os.path.dirname(executable), "julia-py")
+    import sysconfig
+    stempath = sysconfig.get_path('scripts')
+    stempath = os.path.join(stempath, 'julia-py')
     candidates = {os.path.basename(p): p for p in glob.glob(stempath + "*")}
     if not candidates:
         raise RuntimeError(
--- a/src/julia/tools.py
+++ b/src/julia/tools.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, print_function
 
+import sysconfig
 import glob
 import os
 import re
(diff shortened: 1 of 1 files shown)


thautwarm:
This PR fixes `julia-py` for Anaconda Python .

tkf:
Thanks for looking into this. Yes, this does seem like the right way to do it. (Maybe related: #355, #336)


Not sure what is going on with the CI, though.

src/julia/tools.py
```diff
@@ -139,7 +139,9 @@ def julia_py_executable(executable=sys.executable):
     """
     Path to ``julia-py`` executable installed for this Python executable.
     """
-    stempath = os.path.join(os.path.dirname(executable), "julia-py")
+    import sysconfig
```
tkf, line 142: Can you move this up to the beginning of the file?

There are already imports of many packages up there so I don't think we need to lazy-import this particular package.

thautwarm: Yes, and I also remove the default argument. It shall never get used.

tkf:
If you rebase onto `master` the CI should be fixed.

Please pay attention to what Travis job with `TOXENV=style` complains (https://travis-ci.org/github/JuliaPy/pyjulia/jobs/675142992). It nitpicks some styling issues.  If you don't want to wait for travis, you can either

* intall [`tox`](https://tox.readthedocs.io/en/latest/) and run `tox -e style` or
* install [`pre-commit`](https://pre-commit.com/) and run `pre-commit run --all-files`

This PR should be good to go once the CI is all green.

tkf:
Ah, never mind. You just did it :)

thautwarm:
yes, I did see what CI reports haha

tkf:
Thanks!

PR #367, opened 2020-04-15
fix dll deps

thautwarm:
On Windows, with Anaconda Python, when I try

```python
from julia import Julia
jl = Julia()
jl.eval(...)
```
I got an error from `jl_init_with_image__threading`

```
fatal: error thrown and no exception handler available.
<a long trace>
```

I finally find it's due to the lack of loading required dlls like `pcre*.dll`.

This PR fixes those problems, by loading all dynamic libraries in <julia install dir>/bin.

### DIFF
--- a/src/julia/libjulia.py
+++ b/src/julia/libjulia.py
@@ -4,6 +4,7 @@ import ctypes
 import os
 import sys
 from ctypes import POINTER, c_char_p, c_int, c_void_p, pointer, py_object
+from ctypes import cdll, LibraryLoader
 from logging import getLogger  # see `.core.logger`
 
 from .juliainfo import JuliaInfo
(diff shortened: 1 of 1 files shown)


tkf:
Hmm... This doesn't seem right. I think `libjulia` has to be able to handle this. Can you open an issue in JuliaLang/julia?

Also, what happens if you do

```diff
diff --git a/src/julia/libjulia.py b/src/julia/libjulia.py
index 81a8418a..3cd8f220 100644
--- a/src/julia/libjulia.py
+++ b/src/julia/libjulia.py
@@ -311,7 +311,12 @@ class LibJulia(BaseLibJulia):
                 logger.debug("argv[%d] = %r", i, argv[i])

         logger.debug("calling jl_init_with_image(%s, %s)", jl_init_path, sysimage)
-        self.jl_init_with_image(jl_init_path.encode("utf-8"), sysimage.encode("utf-8"))
+        cwd = os.getcwd()
+        try:
+            os.chdir(os.path.dirname(self.libjulia_path))
+            self.jl_init_with_image(jl_init_path.encode("utf-8"), sysimage.encode("utf-8"))
+        finally:
+            os.chdir(cwd)
         logger.debug("seems to work...")

         set_libjulia(self)
```

instead of this PR?

thautwarm:
@tkf it works. so what's happening that you didn't change directory in proper time?

Issue #359, opened 2020-02-10
Building of sysimage fails with pyjulia in user path


oschulz:
When installing pyjulia in my home dir using

    python3 -m pip install --user julia

as recommended in the pyjulia documentation, I'm unable to build a sysimage with

    python3 -m julia.sysimage sys.so

I get a

```
RuntimeError: ``julia-py`` executable is not found for Python installed at /opt/anaconda3/bin/python3
```

I think this is because of

```python
def julia_py(julia, pyjulia_debug, jl_args):
    # ...
    os.environ["_PYJULIA_JULIA_PY"] = julia_py_executable()
```

in "julia_py.py" and 

```python
def julia_py_executable(executable=sys.executable):
    # ...
    stempath = os.path.join(os.path.dirname(executable), "julia-py")
    candidates = {os.path.basename(p): p for p in glob.glob(stempath + "*")}
    # ...
```

in "tools.py".

So `julia_py_executable()` never looks in "$HOME/.local/bin", which is where `python3 -m pip install --user julia` puts `julia-py`. Maybe it could also look for `julia-py` in `site.USER_BASE` and/or `$PATH`?

Kchour:
So I found a workaround by creating a symlink, from where julia-py is installed to `/usr/bin`

To check where julia-py is installed, simply do: `$ type julia-py`

To create the symbolic link: `$ sudo ln -s home/<user>/.local/bin/julia-py /usr/bin/`

Seems to work

grahamgill:
> The best answer I think means getting to know [alternate installation schemes](https://docs.python.org/3/install/index.html#alternate-installation) more intimately. Maybe use `sysconfig.get_scheme_names()` to provide the schemes to check and use `sysconfig.get_paths("scripts",scheme)` for the appropriate scripts directories. 

Here's my attempt to do something along those lines: [tools.py_v0.5.3_diff.txt](https://github.com/JuliaPy/pyjulia/files/4560892/tools.py_v0.5.3_diff.txt)

Is this reasonable?

tkf:
Thanks a lot for giving a shot at this. From a quick look, it seems to be a good direction. It'd be nice if you can open a PR so that it's easier to review/discuss.

tkf:
Also, having a test for this would be great. Though probably the only thing we can do is to use a mock or something.

PR #379, opened 2020-05-04
More robust julia-py discovery (respect alternate installation schemes)

grahamgill:
Addresses JuliaPy/pyjulia#359

### DIFF
--- a/src/julia/tests/test_tools.py
+++ b/src/julia/tests/test_tools.py
@@ -6,74 +6,68 @@ import glob
 
 from julia.tools import julia_py_executable
 
+fake_user = "fa9a5150-8e17-11ea-3f8d-ff1e5ae4a251"
+posix_user_sample_path = os.path.join(os.sep, "home", fake_user, ".local", "bin")
+standard_sample_path = os.path.join(os.sep, "usr", "bin")
+
+
+def get_path_mock_scheme(scheme):
+    # We'll ignore the other alternate install scheme types and just handle "posix_user" and None (standard install).
+    # We use sample scripts paths as reported by sysconfig.get_path("scripts","posix_user") and
+    # sysconfig.get_path("scripts") on a Linux system.
+    assert scheme == "posix_user" or scheme == None
+    if scheme == None:
+        # standard
+        return standard_sample_path
+    else:
+        # posix_user
+        return posix_user_sample_path
+
 
 def julia_py_with_command_extension():
     extension = ".cmd" if platform.system() == "Windows" else ""
     return "julia-py" + extension
 
 
-def glob_mock(path):
-    # we're only handling the case when the glob is ".../julia-py*" or just "julia-py*"
-    # if dirname(path) is empty then return empty list - this is indicator that we shouldn't "find" any files matching the glob
-    return [] if os.path.dirname(path) == "" else [os.path.join(os.path.dirname(path), julia_py_with_command_extension())]
+def glob_mock(path = None):
+    # we're only handling the case when the glob is ".../julia-py*" or nothing
+    # if path is None then return empty list - this is indicator that we don't want to "find" any files matching a glob
+    return [] if path == None else [os.path.join(os.path.dirname(path), julia_py_with_command_extension())]
 
 
 def test_find_julia_py_executable_by_scheme(monkeypatch):
     # could extend this to test different kinds of scheme
-    # right now we just fake the "posix_user" scheme and don't use the rest of get_path_mock_scheme()
-    fake_user = "fa9a5150-8e17-11ea-3f8d-ff1e5ae4a251"
-    fake_folder = "ce410603-12fa-42a2-9a24-e44cd94f6968"
+    # right now we just fake the "posix_user" scheme and standard scheme, giving two paths to look in
 
-    def get_path_mock_scheme(scheme_match, scheme):
-        # sample scripts paths for each scheme type as reported by sysconfig.get_path("scripts",scheme) on a Linux system
-        # sysconfig.get_path("scripts","nt_user") is going to have a different form on Windows, something like "\user\xyz\appdata\..."
-        if scheme == scheme_match:
-            if scheme == "posix_user" or scheme == "osx_framework_user":
-                return os.path.join(os.path.abspath("home"), fake_user, ".local", "bin")
-            elif scheme == "posix_prefix" or scheme == "posix_home":
-                return os.path.join(os.path.abspath("usr"), "bin")
-            elif scheme == "nt":
-                return os.path.join(os.path.abspath("usr"), "Scripts")
-            elif scheme == "nt_user":
-                return os.path.join(os.path.abspath("home"), fake_user, ".local", "Python38", "Scripts")
-            else:
-                return os.path.join(os.path.abspath("tmp"), fake_folder)
-        else:
-            return ""
-
-    monkeypatch.setattr("sysconfig.get_path", lambda x, scheme = None: get_path_mock_scheme("posix_user", scheme))
-    monkeypatch.setattr("glob.glob", lambda x: glob_mock(x))
+    monkeypatch.setattr("sysconfig.get_scheme_names", lambda: ("posix_user",))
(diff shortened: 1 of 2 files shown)


tkf:
Awesome! Thanks for making a PR to tackle one of the really hairy parts of PyJulia.

Did you check it with some platforms? Which ones? It makes sense to note this here because mocking is good for freezing the implementation but not for checking the correctness.

grahamgill:
Unfortunately I'm only in a position to check it works on Linux. Running Manjaro.

tkf:
@thautwarm This touches the part you fixed in #366. Maybe you want to try this branch to make sure it doesn't break in your environment?

src/julia/tools.py
```diff
@@ -140,13 +140,31 @@ def julia_py_executable():
...
+            scripts_paths_failed.append(scripts_path)
+
+    # try standard install scripts path if nothing found in alternate scheme
+    if not candidates:
+        scripts_path = sysconfig.get_path("scripts")
+        stempath = os.path.join(scripts_path, "julia-py")
+        candidates = {os.path.basename(p): p for p in glob.glob(stempath + "*")}
+
     if not candidates:
+        scripts_paths_failed.append(scripts_path)
```
tkf, line 164: It's tricky to figure out `scripts_paths_failed.append(scripts_path)` is not going to add the last `scripts_path` used in `for scheme in sysconfig.get_scheme_names()` (you have to track the state of `candidates`).  Let's rename it to `scripts_paths_tried` and unconditionally append to it in the `for` loop and the first `if not candidates` block.

src/julia/tools.py
```diff
@@ -140,13 +140,31 @@ def julia_py_executable():
...
+        if candidates:
+            break
+        else:
+            scripts_paths_failed.append(scripts_path)
+
+    # try standard install scripts path if nothing found in alternate scheme
+    if not candidates:
+        scripts_path = sysconfig.get_path("scripts")
+        stempath = os.path.join(scripts_path, "julia-py")
+        candidates = {os.path.basename(p): p for p in glob.glob(stempath + "*")}
```
tkf, line 161: Let's factor it out as a function `_julia_py_candidates`.

src/julia/tests/test_tools.py
```diff
@@ -0,0 +1,79 @@
...
+    monkeypatch.setattr("sysconfig.get_path", lambda x, scheme = None: get_path_mock_not_found())
+    monkeypatch.setattr("glob.glob", lambda x: glob_mock(x))
+
+    try:
+        julia_py_executable()
+        notfound = False
+    except RuntimeError as exc:
+        notfound = "``julia-py`` executable is not found" in exc.args[0]
+
+    assert notfound
```
tkf, line 79: Why not https://docs.pytest.org/en/latest/reference.html#pytest-raises ?

src/julia/tools.py
tkf: Actually, I think it'd be better to do

```python
scripts_paths = [sysconfig.get_path("scripts", scheme) for scheme in sysconfig.get_scheme_names()]
scripts_paths.append(sysconfig.get_path("scripts"))
```

and loop over `scripts_paths`.

src/julia/tests/test_tools.py
```diff
@@ -0,0 +1,79 @@
...
+    # right now we just fake the "posix_user" scheme and don't use the rest of get_path_mock_scheme()
+    fake_user = "fa9a5150-8e17-11ea-3f8d-ff1e5ae4a251"
+    fake_folder = "ce410603-12fa-42a2-9a24-e44cd94f6968"
+
+    def get_path_mock_scheme(scheme_match, scheme):
+        # sample scripts paths for each scheme type as reported by sysconfig.get_path("scripts",scheme) on a Linux system
+        # sysconfig.get_path("scripts","nt_user") is going to have a different form on Windows, something like "\user\xyz\appdata\..."
+        if scheme == scheme_match:
+            if scheme == "posix_user" or scheme == "osx_framework_user":
+                return os.path.join(os.path.abspath("home"), fake_user, ".local", "bin")
```
tkf, line 32: I'm not sure why you are using `os.path.abspath`.  Maybe you mean `os.path.join(os.sep, "home", fake_user, ...)`?

src/julia/tests/test_tools.py
```diff
@@ -0,0 +1,79 @@
...
+
+def glob_mock(path):
+    # we're only handling the case when the glob is ".../julia-py*" or just "julia-py*"
+    # if dirname(path) is empty then return empty list - this is indicator that we shouldn't "find" any files matching the glob
+    return [] if os.path.dirname(path) == "" else [os.path.join(os.path.dirname(path), julia_py_with_command_extension())]
+
+
+def test_find_julia_py_executable_by_scheme(monkeypatch):
+    # could extend this to test different kinds of scheme
+    # right now we just fake the "posix_user" scheme and don't use the rest of get_path_mock_scheme()
```
tkf, line 23: Let's not add code that we don't use. I think you can just patch `get_scheme_names` and make sure that it always returns `"posix_user"` and then `assert` it in `get_path_mock_scheme`.

src/julia/tests/test_tools.py
```diff
@@ -0,0 +1,79 @@
...
+                return os.path.join(os.path.abspath("usr"), "Scripts")
+            elif scheme == "nt_user":
+                return os.path.join(os.path.abspath("home"), fake_user, ".local", "Python38", "Scripts")
+            else:
+                return os.path.join(os.path.abspath("tmp"), fake_folder)
+        else:
+            return ""
+
+    monkeypatch.setattr("sysconfig.get_path", lambda x, scheme = None: get_path_mock_scheme("posix_user", scheme))
+    monkeypatch.setattr("glob.glob", lambda x: glob_mock(x))
```
tkf, line 45:
```suggestion
    monkeypatch.setattr("glob.glob", glob_mock)
```

src/julia/tests/test_tools.py
grahamgill: Great, sure. I wanted to illustrate what's available in case needed. I've cut it down.

src/julia/tests/test_tools.py
grahamgill: I called `os.path.abspath("home")` etc. to do just that, get `/home` to start the path off without depending on the particular symbol on *nix and MacOS, but wasn't sure whether Windows would want to stick in a drive letter at the start also. Anyway will just use `os.sep` .

src/julia/tests/test_tools.py
grahamgill: Because only now do I know it exists! Thanks.

src/julia/tools.py
grahamgill: OK yes, have collapsed the two sections into one as suggested.

src/julia/tools.py
grahamgill: `scripts_paths_tried` is redundant with the suggestion to gather both alternate and standard paths together and then loop over `scripts_paths`. If we don't find any `candidates` then we've already tried all `scripts_paths` and we can report all `scripts_paths` as the ones we tried.

thautwarm:
@grahamgill could you compare this with the specifications? Maybe you'd like to check https://github.com/JuliaPy/pyjulia/pull/366
Is it really necessary to make a big change to the existing codebase?
According to [alternate-installation-the-user-scheme](https://github.com/python/cpython/blob/master/Doc/install/index.rst#alternate-installation-the-user-scheme), we just need to get variables from this module:
https://github.com/python/cpython/blob/master/Lib/sysconfig.py

tkf:
@grahamgill I added a bit of style tweaks. Next time, please use `pre-commit run --all-files` or `tox -e style` to resolve what CI complains.

grahamgill:
@tkf Yes I had just started trying to figure out which order of `import`s would satisfy `tox -e style` earlier today but had to put that aside for later.

OK I see the standard must be PEP 8.

tkf:
@grahamgill Thanks for the patch!

@thautwarm I'll just go ahead and merge this PR. Let me know if you find anything strange in Windows with PyJulia `master`.

PR #380, opened 2020-05-06
create pyjulia sysimage on top of existing custom sysimage

grahamgill:
I have a Julia package for which I build a custom sysimage using PackageCompiler. I need to make this package and the performance gains of the custom sysimage available in python.

I've done this in PyJulia by building its custom sysimage on top of a provided sysimage, using the `base_sysimage` parameter of `create_sysimage` in PackageCompiler. Others may find this useful.


tkf:
Can you add a test like this?

https://github.com/JuliaPy/pyjulia/blob/e4c51b2a8fd0608a1717c1f6e929a7485a2badd0/src/julia/tests/test_sysimage.py#L9-L40

You can just use the default system image as `base_sysimage`. I think you can get the path of the default system image from `juliainfo.sysimage`.

grahamgill:
One caveat is that `base_sysimage` is not documented for `create_sysimage` in [PackageCompiler.jl](https://github.com/JuliaLang/PackageCompiler.jl/blob/master/src/PackageCompiler.jl). But looking at the code for `create_sysimage` it's straightforward to see that the `base_sysimage` parameter is used to provide a base sysimage (of course! :smile: ) on which to build incrementally, instead of the default system image.

tkf:
To be extra careful, it'd be nice if you can open a PR in PackageCompiler.jl to add it to the documentation.  This would put `base_sysimage` in the stable API of PackageCompiler.jl so that we can safely rely on it (otherwise it may be renamed or removed at some point).

grahamgill:
@tkf Added as you've suggested. I added the test to `test_sysimage.py` and factored out common components from the two tests.

I wasn't able to test on my laptop with `pytest` since there doesn't seem to be a `pytest.ini` file defining `@pytest.mark.julia`, or if there's some other mechanism then I don't know how that works. `pytest` skipped the two tests with an UnknownMarkWarning.

I was able to check the two tests directly in a python console.

> Can you add a test like this?
> 
> https://github.com/JuliaPy/pyjulia/blob/e4c51b2a8fd0608a1717c1f6e929a7485a2badd0/src/julia/tests/test_sysimage.py#L9-L40
> 
> You can just use the default system image as `base_sysimage`. I think you can get the path of the default system image from `juliainfo.sysimage`.

tkf:
Thanks, I like the way the tests are organized :+1:

> since there doesn't seem to be a `pytest.ini` file defining `@pytest.mark.julia`

This is because you need to enable PyJulia plugin for pytest (see `src/julia/runtests.py` for how it's done).

The easiest way to run tests is to use `tox`.  Note that sysimage test is run only on CI by default so you need to do something like `CI=true tox -- -k test_sysimage`.   If you want try it outside `tox`, you can use `CI=true python -m julia.runtests -- -k test_sysimage`.

See also: https://pyjulia.readthedocs.io/en/latest/testing.html

tkf:
Oops. I probably should have waited for the PackageCompiler.jl PR.

grahamgill:
@tkf Great, thanks for the tips.

> Oops. I probably should have waited for the PackageCompiler.jl PR.

JuliaLang/PackageCompiler.jl#388 explains why Julia nightly build Travis tests aren't passing on my PR JuliaLang/PackageCompiler.jl#401 (which only changes a doc string).

KristofferC:
I don't understand the need for `base_sysimage` over using `incremental=false`.

PR #383, opened 2020-05-18
Warning

PallHaraldsson:


PallHaraldsson:
FYI: I've been bitten by, while being able to run julia, $ whereis julia
gets me nothing, and thus e.g. this package didn't work. The full path requirement is likely a security issue, and took me a while to realize.

Also:

$ snap install julia

sort of works, that is most of the time, until e.g. using your package (or was it PyCall? probably it needs a warning too).

PallHaraldsson:
FYI: This package doesn't work: https://github.com/emmettgb/PyLathe/blob/master/Lathe/__init__.py

Nor his other package for me (both using pyjulia):

https://github.com/emmettgb/TopLoader

```
>>> import TopLoader
TopLoader V 0.0.1
Happy Loading!
-----------------
  File "/usr/local/lib/python3.6/dist-packages/TopLoader/__init__.py", line 5, in <module>
    from julia import Pkg
ImportError: cannot import name 'Pkg'
```

emmettgb:
Hmm there is a known bug on Ubuntu, are you using Ubuntu or Debian by chance?

emmettgb:
The issue is that Snap and/or apt puts your Julia binaries into a different location than PyJulia looks for. I've encountered the same issue on a Chromebook with a Debian Docker image.... Fortunately, I daily drive Fedora, where this issue is non-existent. I have yet to come up with a good solution, but I'm guessing there is an environmental variable you could possibly change.

PallHaraldsson:
Yes, using Ubuntu. If you look at the latter comment, it might be that?

PallHaraldsson:
For the both of you, while I got pyjulia working previously, and as late as today, I'm a bit mistified on why it stopped working, and I do not get further:

```
$ python3
>>> import julia
>>> julia.install(julia='/home/pharaldsson_sym/julia-1.4.0/bin/julia') # for any or no path known to work
AttributeError: 'module' object has no attribute 'install'

$ whereis julia
julia: /home/pharaldsson_sym/julia-1.5-ea669c3d3e/bin/julia
```

tkf:
> `AttributeError: 'module' object has no attribute 'install'`

This is strange. `julia.install` should be there if `import julia` succeeds (unless this PyJulia is very old). Maybe try `dir(julia)` when it happens and see if `install` is listed.

PallHaraldsson:
>This is strange. julia.install should be there if import julia succeeds (unless this PyJulia is very old). 

Strange right, it's not just me. Worked previously in he day, with I guess latest (and still):

```
$ pip3 install julia
Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: julia in /home/pharaldsson_sym/.local/lib/python3.6/site-packages (0.5.3)

$ python3
Python 3.6.9 (default, Apr 18 2020, 01:56:04) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> dir(julia) 
NameError: name 'julia' is not defined
>>> import julia
>>> dir(julia) 
['Julia', 'JuliaError', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'core']
>>> 
```

tkf:
If your PyJulia does not even have `__version__`, it's probably very old. You need to update it.  Maybe try `python3 -m pip install --upgrade julia`?

PallHaraldsson:
At least the warning for core.py could be helpful. Maybe this got forgotten as "21 expected" checks? I've never seen such before, and this has been stalled for long...

PR #407, opened 2020-08-03
Use `Base.require` instead of `using` in more places

tkf:
## Commit Message
Use `Base.require` instead of `using` in more places (#407)

This way, PyJulia can work in more strict setup (e.g., customized
`JULIA_LOAD_PATH`).


tkf:
@mergifyio refresh

PR #420, opened 2020-09-14
Fix test__using_default_setup with PYJULIA_TEST_RUNTIME

tkf:
## Commit Message
Fix test__using_default_setup with PYJULIA_TEST_RUNTIME (#420)

Before this patch,

    PYJULIA_TEST_RUNTIME=dummy tox -- --no-julia -k test__using_default_setup

did not work (which came up when I was writing
https://github.com/tkf/PyJuliaTester.jl).  It is fixed by removing
`PYJULIA_TEST_RUNTIME` in `test__using_default_setup`.

### DIFF
--- a/.github/workflows/testtests.yml
+++ b/.github/workflows/testtests.yml
@@ -1,0 +1,21 @@
+name: Test tests
+
+on:
+  push:
+    branches:
+      - master
+    tags: '*'
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  test__using_default_setup:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup python
+        uses: actions/setup-python@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/testtests.yml
```diff
@@ -0,0 +1,21 @@
...
+  test__using_default_setup:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup python
+        uses: actions/setup-python@v1
+      - run: python -m pip install --upgrade tox
+      - run: python -m tox -- --no-julia -k test__using_default_setup
+        env:
+          PYJULIA_TEST_RUNTIME: dummy
```
tkf, line 21: Repro works: https://github.com/JuliaPy/pyjulia/pull/420/checks?check_run_id=1109711015#step:5:91

PR #422, opened 2020-09-15
allow building sysimage with custom PyCall

marius311:
The current sysimage just installs the latest PyCall from the registry. This adds a `--pycall-env` option which lets you bake in an existing one from any project you want, 

```shell
$ python -m julia.sysimage --pycall-env . sys.so # take PyCall from the project in the current folder
```

I found this useful in my debugging  / developing, and I think it'd be nice to have in the official repo.

### DIFF
--- a/src/julia/compile.jl
+++ b/src/julia/compile.jl
@@ -1,4 +1,4 @@
-compiler_env, script, output, base_sysimage = ARGS
+compiler_env, pycall_env, script, output, base_sysimage = ARGS
 
 if VERSION < v"0.7-"
     error("Unsupported Julia version $VERSION")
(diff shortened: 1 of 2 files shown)


tkf:
Thanks! Yes, it'd be nice to have.

Can you add a test for this? You can put it in https://github.com/tkf/pyjulia/blob/pyjuliatester/src/julia/tests/test_sysimage.py

Also, I wonder if it makes sense to call `--pycall-env` something else. For example, it might be useful to list, e.g., PyPlot and Plots in such an environment (even though currently they are not get loaded). Maybe `--sysimage-env`? Though `--pycall-env` kinda still makes sense since we always have PyCall in it.

Other than that, it LTGM!

mkitti:
Did we ever add a test?

Issue #437, opened 2021-04-05
Pyjulia not working on Julia 1.6


lvzhibai:
When I try to install and work with Python3.9.3 with Julia 1.6, the command
> julia.install()

works well, but later
> from julia import Base # or `Pkg`, `Main` ..etc

leads to segmentation fail, and python exits without other error messages inside terminal.

Then I reinstalled julia 1.5, all commands works well.

emanuellima1:
I tried with Python 3.8.6 and Julia 1.6 on Kubuntu 20.10 and everything worked fine. I did:

```python
import julia
julia.install()
jl = julia.Julia(compiled_modules=False)
from julia import Base
Base.sind(90)
```

On the Python REPL.

ampbj:
I have the same issue:
MacOS Big Sur.
Python 3.8.8 [Clang 10.0.0 ] :: Anaconda, Inc. on darwin
julia version 1.6.0

JackDunnNZ:
For reference, there is a similar problem for 1.6 with JuliaCall in R on macOS: https://github.com/Non-Contradiction/JuliaCall/issues/164

So it seems to be a bug in Julia, the issue is being tracked here: https://github.com/JuliaLang/julia/issues/40246

KwatME:
Same problem here. I'm currently using 1.5.3 until this problems fixed.

fchorney:
I posted this in https://github.com/JuliaLang/julia/issues/40246, but I'm going to post it here as well to get a bit more exposure.

----

So I've been doing some testing trying to get macOS 11.4, Python 3.9, and Julia 1.6 working with pyjulia 0.5.6, and I think I found a way to get it to work somewhat.

Some notes:
- I am using the Julia 1.6 macOS binary downloaded from julialang.org which exists in `/Applications/Julia-1.6.app`.
- I have symlinked the julia binary to my `/usr/local/bin` like so `ln -s /Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia /usr/local/bin/julia1.6` and then whenever I change my default julia I symlink it again like so: `ln -s /usr/local/bin/julia1.6 /usr/local/bin/julia`
- I am doing this in a python 3.9 venv 

So my first few observations.
- I had tried this with the Julia 1.7 beta and noticed a few extra debug lines.
[...]

PR #440, opened 2021-04-14
Fix julia library jl_init_with_image bug

southpawac:
A possible solution to #437 

`jl_init_with_image` and `jl_init_with_image__threading` are no longer exposed functions in the Julia 1.6 library, whereas `jl_init__threading` seems to work.


iamed2:
AFAICT `jl_init_with_image__threading` is still exposed, just like `jl_init__threading`

southpawac:
You're right. 
This is just a temporary workaround until `jl_init_with_image__threading` is fixed.

PR #445, opened 2021-06-16
try `PYTHONPATH` in `julia_py_executable` to find julia-py

manuelbb-upb:
I propose to test the path(s) set in the environment variable `PYTHONPATH` to be considered as candidates for the `julia-py` script.

Consider the case when `pyjulia` has been installed in a non-standard location, e.g., via 
```
pip install --target=/some/dir julia
```
(The `julia-py` script is put into `/some/dir/bin`.)

We can easily get most of the functionality by doing something like 
```python
import sys
sys.path.append("/some/dir")
```
However, compiling a sysimage is not possible:
1) When trying to call the `build_sysimage` function from within a python script, then the `julia-py` script cannot be found.
2) When trying  
    ```
    python -m julia.sysimage sys.img
    ```
    in the console, then the module is not found, because Python does not know about "/some/dir".

Both approaches can be made to work with the proposed solution:
1) Set the env var with `os.environ["PYTHONPATH"] = "/some/dir"` and then call `build_sysimage`.
2) Do `export PYTHONPATH=/some/dir` and then compile the sysimage from the console.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -29,7 +29,7 @@ jobs:
         if: >-
           (github.ref == 'refs/heads/release/test') ||
           (github.ref == 'refs/heads/release/main')
-        uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29  # v1.4.2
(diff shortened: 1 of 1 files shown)


mkitti:
oops

PR #481, opened 2021-12-29
Fix failure setting optimize option (#480)

### LINKED ISSUE #480: Failure when setting optimize option
This is because pyjulia uses ArgumentsParser to parse --home & --sysimage, but it get the optimize value as an int instead of a string.

I'll submit a PR

dpinol:
Fixes #480

### DIFF
--- a/src/julia/tests/test_juliaoptions.py
+++ b/src/julia/tests/test_juliaoptions.py
@@ -11,7 +11,7 @@ from julia.core import JuliaOptions
     (dict(compiled_modules="no"), ["--compiled-modules", "no"]),
     (dict(depwarn="error"), ["--depwarn", "error"]),
     (dict(sysimage="PATH"), ["--sysimage", "PATH"]),
-    (dict(bindir="PATH"), ["--home", "PATH"]),
+    (dict(optimize=3), ["--optimize", "3"]),
 ])
 # fmt: on
 def test_as_args(kwargs, args):
--- a/src/julia/tests/test_juliaoptions.py
+++ b/src/julia/tests/test_juliaoptions.py
@@ -11,6 +11,7 @@ from julia.core import JuliaOptions
     (dict(compiled_modules="no"), ["--compiled-modules", "no"]),
     (dict(depwarn="error"), ["--depwarn", "error"]),
     (dict(sysimage="PATH"), ["--sysimage", "PATH"]),
+    (dict(bindir="PATH"), ["--home", "PATH"]),
     (dict(optimize=3), ["--optimize", "3"]),
 ])
 # fmt: on
--- a/src/julia/options.py
+++ b/src/julia/options.py
@@ -134,7 +134,7 @@ class JuliaOptions(object):
     compile = Choices("compile", yes_no_etc("all", "min"))
     depwarn = Choices("depwarn", yes_no_etc("error"))
     warn_overwrite = Choices("warn_overwrite", yes_no_etc())
-    optimize = Choices("optimize", dict(zip(range(4), range(4))))
+    optimize = Choices("optimize", dict(zip(range(4), map(str, range(4)))))
     inline = Choices("inline", yes_no_etc())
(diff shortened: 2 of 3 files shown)


mkitti:
We would need to get these tests passing

mkitti:
@MilesCranmer Clone the branch, get the tests to pass, and submit a new PR pinging me. Then I will merge it.

dpinol:
logs have expired. Can sb rerun the tests?

mkitti:
Considering rebasing

dpinol:
@mkitti done

mkitti:
Hmm, that did not seem to fix testing.

MilesCranmer:
It seems unrelated to this PR - the issue seems to be because the CI is trying to download a pip package from an HTTP URL instead of HTTPS, and urllib no longer allows non-SSL requests:
```
      Downloading http://pypi.python.org/packages/source/d/distribute/distribute-0.6.27.tar.gz
```
I think that's why the Python 2.7 version works - since the old urllib hasn't deprecated non-SSL.

MilesCranmer:
FYI @dpinol once #508 is merged this should work

dpinol:
@MilesCranmer rebased. thanks

MilesCranmer:
@mkitti any idea how to change the actions for “Required statuses must pass before merging”? It seems like they are for a different CI naming scheme maybe.

MilesCranmer (approved):
Looks good to me.

src/julia/tests/test_juliaoptions.py
```diff
@@ -11,7 +11,7 @@
     (dict(compiled_modules="no"), ["--compiled-modules", "no"]),
     (dict(depwarn="error"), ["--depwarn", "error"]),
     (dict(sysimage="PATH"), ["--sysimage", "PATH"]),
-    (dict(bindir="PATH"), ["--home", "PATH"]),
```
mkitti, line 14: What happened to `bindir`?

mkitti:
Status checks should be functional now

src/julia/tests/test_juliaoptions.py
mkitti: Other than this merge when ready.

MilesCranmer: Is bindir an option in Julia anymore? I don't see it in the `--help` menu.

MilesCranmer:
Could this one be merged?

src/julia/tests/test_juliaoptions.py
mkitti: https://github.com/JuliaLang/julia/blob/master/src/jloptions.c

MilesCranmer: Not sure, I can't even use it locally:

```bash
julia --bindir=$PATH
# ERROR: unknown option `--bindir
```

MilesCranmer: I think it's replaced by `--home`?

MilesCranmer: Oh yeah, there it is:

```python
    def cli_argument_name(self):
        name = {"bindir": "home"}.get(self.name, self.name)
```

So this should not be removed @dpinol

src/julia/tests/test_juliaoptions.py
```diff
@@ -11,7 +11,7 @@
...
+    (dict(optimize=3), ["--optimize", "3"]),
```
mkitti, line 14:
```suggestion
    (dict(bindir="PATH"), ["--home", "PATH"]),
    (dict(optimize=3), ["--optimize", "3"]),
```

src/julia/tests/test_juliaoptions.py
mkitti: I readded it.

MilesCranmer:
Awesome! 🚀

PR #524, opened 2023-02-28
Bump version to 0.6.1 with Julia 1.9 fix

MilesCranmer:
It turns out I can't push to master. @mkitti do you want to merge this and also create a release tagged at the merge commit?


mkitti:
Sure. Let CI run its course, and then auto merge should work.

MilesCranmer:
Thanks. FYI It still says the Julia 1.0 jobs are required, not sure why (i.e., auto-merge won't work)

PR #530, opened 2023-05-05
adding the trace-compile option to the julia initialization

rdyro:
It'd be nice to be able to trace the compilation performed by Julia called via PyJulia. The main use case is for the precompilation of a Julia library triggered primarily through their Python interface.

The change is a single line adding the `trace-compile` option.

### DIFF
--- a/src/julia/tests/test_trace_compile.py
+++ b/src/julia/tests/test_trace_compile.py
@@ -25,5 +25,9 @@ def test_trace_file_created(tmpdir):
     # check whether the sin precompilation directive is included in the file
     trace_compile_content = Path(trace_compile_path).read_text().strip()
     lines = [x for x in trace_compile_content.split("\n") if len(x) > 0]
-    expected_precompile_line = r"precompile\(Tuple\{typeof\([A-Za-z]+\.sin\), Float64\}\)"
-    assert any([re.match(expected_precompile_line, x) is not None for x in lines])
+    expected_precompile_line = (
+        r"precompile\(Tuple\{typeof\([A-Za-z]+\.sin\), Float64\}\)"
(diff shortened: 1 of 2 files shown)


mkitti:
Thanks. Could you work on updating the tests?

rdyro:
I'll take a look and push another commit soon, thanks!

rdyro:
I fixed all errors in existing tests and I added a new test specifically for the new `trace_compile` option.

If you can take a look now, that'd be great. @mkitti

src/julia/tests/test_trace_compile.py
```diff
@@ -0,0 +1,26 @@
...
+    jl = Julia(trace_compile="{}")
+
+    from julia import Main
+    Main.sin(2.0)
+    """.format(
+            str(trace_compile_path)
+        )
+    )
+    assert (trace_compile_path).exists()
+    assert len(trace_compile_path.read_text().strip()) > 0
```
MilesCranmer, line 26: Is there some compilation string that you know should be printed to this file? If so, could you test that it actually shows up there?

MilesCranmer (changes requested):
Looks good. I have one suggestion for the test to make sure it's actually working (and not just printing random stuff to the file). Sorry for taking so long to look at this!

src/julia/tests/test_trace_compile.py
rdyro: Yes, that's a great idea! I added a check for the precompilation directive of the `sin` function we invoke to test precompilation.

rdyro:
I think with the additional test, this should look better now. Thanks for taking a look @MilesCranmer

PR #539, opened 2023-11-03
Set `compiled_modules=False` automatically

### LINKED ISSUE #487: Turn on compiled_modules=False automatically?
Could compiled_modules=False be turned on automatically if the user does not have a dynamically linked libpython? As far as I can tell this doesn't present too big of a problem for a user to have the precompilation cache turned off.

https://github.com/JuliaPy/pyjulia/blob/56a739126d3fed0d24e548069ff216fea78fbe1b/src/julia/core.py

This is currently an error, but I think it should instead be a warning to just let the user know that precompilation won't work. I think it should automatically re-run with `Julia(..., compiled_modules=False)` if `compiled_modules=True` does not work.

I do this automatically in my library which uses PyJulia: https://github.com/MilesCranmer/PySR/blob/57de9547c4d09ddada4249aa25b541f835fa04b3/pysr/sr.py#L965-L1003, so that users don't have to worry about python versions, but I think it should also be turned on in PyJulia itself.

Moelf:
fix #487

@MilesCranmer

### DIFF
--- a/src/julia/core.py
+++ b/src/julia/core.py
@@ -485,7 +485,7 @@ class Julia(object):
                 is_compatible_python
                 or use_custom_sysimage
             ):
-                if options.compiled_modules == "yes":
+                if options.compiled_modules in (True, "yes"):
                     raise UnsupportedPythonError(jlinfo)
                 else:
                     warnings.warn(
--- a/src/julia/core.py
+++ b/src/julia/core.py
@@ -482,11 +482,17 @@ class Julia(object):
             logger.debug("use_custom_sysimage = %r", use_custom_sysimage)
             logger.debug("compiled_modules = %r", options.compiled_modules)
             if not (
-                options.compiled_modules == "no"
-                or is_compatible_python
+                is_compatible_python
                 or use_custom_sysimage
             ):
-                raise UnsupportedPythonError(jlinfo)
+                if options.compiled_modules == "yes":
+                    raise UnsupportedPythonError(jlinfo)
+                else:
+                    warnings.warn(
+                            "Statically linked Python interpreter detected, setting `compiled_modules=False` automatically."
+                            )
+                    options.compiled_modules = "no"
+
 
             self.api.init_julia(options)
 
--- a/src/julia/core.py
+++ b/src/julia/core.py
@@ -487,6 +487,8 @@ class Julia(object):
             ):
                 if options.compiled_modules in (True, "yes"):
                     raise UnsupportedPythonError(jlinfo)
+                elif options.compiled_modules in (False, "no"):
+                    pass
                 else:
                     warnings.warn(
                             "Statically linked Python interpreter detected, setting `compiled_modules=False` automatically."
--- a/src/julia/options.py
+++ b/src/julia/options.py
@@ -165,7 +165,7 @@ class JuliaOptions(object):
 
     sysimage = String("sysimage")
     bindir = String("bindir")
-    compiled_modules = Choices("compiled_modules", yes_no_etc())
+    compiled_modules = Choices("compiled_modules", yes_no_etc(), "auto")
     compile = Choices("compile", yes_no_etc("all", "min"))
     depwarn = Choices("depwarn", yes_no_etc("error"))
     warn_overwrite = Choices("warn_overwrite", yes_no_etc())
--- a/src/julia/options.py
+++ b/src/julia/options.py
@@ -165,7 +165,7 @@ class JuliaOptions(object):
 
     sysimage = String("sysimage")
     bindir = String("bindir")
-    compiled_modules = Choices("compiled_modules", yes_no_etc(), "auto")
+    compiled_modules = Choices("compiled_modules", yes_no_etc("auto"))
     compile = Choices("compile", yes_no_etc("all", "min"))
     depwarn = Choices("depwarn", yes_no_etc("error"))
     warn_overwrite = Choices("warn_overwrite", yes_no_etc())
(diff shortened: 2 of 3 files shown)


src/julia/core.py
```diff
@@ -482,11 +482,17 @@ def __init__(self, init_julia=True, jl_init_path=None, runtime=None,
             logger.debug("use_custom_sysimage = %r", use_custom_sysimage)
             logger.debug("compiled_modules = %r", options.compiled_modules)
             if not (
-                options.compiled_modules == "no"
-                or is_compatible_python
+                is_compatible_python
                 or use_custom_sysimage
             ):
-                raise UnsupportedPythonError(jlinfo)
+                if options.compiled_modules == "yes":
```
MilesCranmer, line 488:
```suggestion
                if options.compiled_modules in (True, "yes"):
```

src/julia/core.py
```diff
@@ -482,11 +482,17 @@ def __init__(self, init_julia=True, jl_init_path=None, runtime=None,
...
+                    raise UnsupportedPythonError(jlinfo)
+                else:
```
MilesCranmer, line 490:
```suggestion
                elif options.compiled_modules in (False, "no"):
                    pass
                else:
```

MilesCranmer:
Also I think the default `compiled_modules` should then be `"auto"`. Otherwise the default will trigger errors anyways.

Moelf:
it's "auto" now

src/julia/tests/test_juliaoptions.py
```diff
@@ -17,7 +17,7 @@
     (dict(min_optlevel=2), ["--min-optlevel=2"]),
     (dict(threads="auto", optimize=3), ["--optimize=3", '--threads=auto']),
     (dict(optimize=3, threads="auto"), ["--optimize=3", '--threads=auto']),  # passed order doesn't matter
-    (dict(compiled_modules=None, depwarn="yes"), ["--depwarn=yes"]),
+    (dict(compiled_modules="auto", depwarn="yes"), ["--depwarn=yes"]),
```
MilesCranmer, line 20: I think this is just the test file, no? I think it needs to be set here: https://github.com/JuliaPy/pyjulia/blob/69b734f117e489904a07d20f46c8f96b56577779/src/julia/options.py#L166-L176

Moelf:
I think that's how you set a default? I don't see any example using that final argument of `Choices` :(

src/julia/options.py
```diff
@@ -165,7 +165,7 @@ class JuliaOptions(object):
 
     sysimage = String("sysimage")
     bindir = String("bindir")
-    compiled_modules = Choices("compiled_modules", yes_no_etc())
+    compiled_modules = Choices("compiled_modules", yes_no_etc(), "auto")
```
MilesCranmer, line 168:
```suggestion
    compiled_modules = Choices("compiled_modules", yes_no_etc("auto"))
```

Moelf: https://github.com/JuliaPy/pyjulia/blob/bfca4a2043229a0bee8d4ea41bfd31e10904e532/src/julia/options.py#L101-L105

I thought that just adds an entry into the dict `"auto": "auto"`

Moelf:
How many CIs are expected to pass?

MilesCranmer:
All the tests that run should pass. I think the "Required" ones should be turned off as they don't run anymore.

MilesCranmer:
Not sure what the git errors are from but this error seems to be real: https://github.com/JuliaPy/pyjulia/actions/runs/6748661679/job/18347465730?pr=539

mkitti:
This would be breaking, so we're going to need a version bump.

Moelf:
it's not breaking though. Before it errors, now it warns. Performance regression is not breaking

MilesCranmer:
> Before it errors, now it warns

So one example is PySR has logic for catching the default unsupported python error and then turning off compiled_modules with some extra instructions on how to install pyenv. Not sure what other packages may be affected.

Moelf:
that's precisely why we want this cuz you're already doing it -- again, being able to use something (in both direction) is better than nothing

MilesCranmer:
Oh yeah I completely agree! Just was saying that this is worth a 0.6 -> 0.7 since it is a change in behavior and would require downstream changes to accommodate

Moelf:
would it? it might make your switching logic in pysr dead code, but it shouldn't break it?

MilesCranmer:
Right but other packages may use it differently

Moelf:
I feel that relying on error and do something non-trivial (other than recovering like pysr) is a bit weird, but okay yeah technically it's possible

mkitti:
You've changed the default behavior with respect to `compiled_modules`. That's fine, but I may prioritize non-breaking changes for merging and release.

Moelf:
I still don't see how it's breaking, but I don't really care about the semantics debate. We should just try to get this out

Moelf:
is our CI broken?

MilesCranmer:
@Moelf I think I got it working - was just a test for the docs which failed, because they didn't include the new `"auto"` option.

Could you perhaps add a test with a static library so we can verify it works? For example like this: https://github.com/MilesCranmer/PySR/blob/f178813c1f8aaefd2a19006dd4d24b645d9eaf76/.github/workflows/CI.yml#L110-L138 – because conda has a static python library so would normally fail this test.

Moelf:
Can do, but why do half of the tests still failing?

MilesCranmer:
The failures look real. This is the test name: `test_compatible_exe.py::test_incompatible_python`. I guess they just need to be updated to the new behavior.

MilesCranmer:
See, for example:

https://github.com/JuliaPy/pyjulia/blob/71b81b68177d1c068e0be88f130d52f70aacf98e/src/julia/tests/test_compatible_exe.py#L192

MilesCranmer:
You could add `compiled_modules=True` to the line

```
Julia(runtime=os.getenv("PYJULIA_TEST_RUNTIME"), debug=True)
```

and I think that should solve it?

PR #538, opened 2023-10-30
Fix PyJulia on Python 3.12

MilesCranmer:
cc @mkitti do you think you could give this a look over? It still seems like there are some issues even though `from julia import Main` doesn't blow up... Any help appreciated.

This updates the importlib API according to what was noticed in https://github.com/JuliaPy/pyjulia/issues/509 (which were giving deprecation warnings, we simply didn't update in time).

This is a lot of guess work about the importlib API: https://docs.python.org/3/library/importlib.html, so I'd appreciate if someone could give a close reading.

I found this stackoverflow answer to be particularly helpful: https://stackoverflow.com/a/43573798/2689923

cc @IlyaOrson this should finally fix the PySR issue you saw (the proper way!)

PR #541, opened 2023-11-05
Update setup actions and get working on 3.12

MilesCranmer:
This PR tries to fix the CI setting issues in #538, #539, #540. Those PRs can then set this branch to their base.

It also merges the Python 3.12 fix from #538.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,10 +22,9 @@ jobs:
           - '3.9'
           - '3.12'
         julia-version:
-          - '1.4'
-          - '1.7'
-          - '1.8'
+          - '1.6'
           - '1.9'
+          - '1'
         exclude:
           - os: ubuntu-latest
             architecture: x86
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,9 +59,9 @@ jobs:
       jl${{ matrix.julia-version }}
       ${{ matrix.os }} ${{ matrix.architecture }}
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v4
       - name: Setup python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
           architecture: ${{ matrix.architecture }}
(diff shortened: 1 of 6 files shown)


MilesCranmer:
@mkitti @moelf any idea what this error is from? I haven't seen it before in the CI.

```
ERROR: Unable to dlopen(cxxpath) in parent!
Message: /lib/x86_64-linux-gnu/libstdc++.so.6�: cannot open shared object file: No such file or directory
```

does it require a specific additional library installed?

I wonder if it's related to that issue we debugged a while back in PyCall.jl: https://github.com/JuliaPy/PyCall.jl/pull/1040

Moelf:
is it possible it's related to Github updated Ubuntu base version? But I don't understand why other Python and Julia combo don't error

mkitti:
We should add Julia 1.9 tests. I would be in favor of dropping everything else except Julia 1.6 since it is LTS.

Julia 1.4, Julia 1.7, and Julia 1.8 are no longer supported.

MilesCranmer:
What apt library should I try to add to get the missing C++ library?

MilesCranmer:
I've updated the CI to only test Julia 1.6, 1.9, and 1 (=1.10 currently).

It looks like that libstdc++ issue was some unrelated issue with the action runner, rather than the environment being set up incorrectly. 

The remaining issue is a randomly-occurring "error 245" which apparently is a segfault. But this doesn't appear to be caused by any of the changes in this PR so I suggest we temporarily ignore it and merge, so that we can have a working Python 3.12 version.

Is that okay with you @mkitti? I am unable to merge as I don't have the access rights (since it marks some tests as "Required" even though they aren't ran anymore). So once you are satisfied would it be possible for you to merge?

mkitti:
This does look better. Let me re-run that one Ubuntu test.

MilesCranmer:
Awesome, thanks!

MilesCranmer:
How do we release the version? I see there is an action with a workflow_dispatch... Do I just trigger that?

MilesCranmer:
Oh wait I think I figured it out. Let me try...

PR #546, opened 2023-12-30
Don't fail CI on failing Windows 32-bit

MilesCranmer:
cc @mkitti @Moelf

This changes the main test suite to not fail the entire CI (with the red X) if the 32-bit windows test fails. This action runner seems extremely flaky, and is currently broken, and I am worried that the red X loses its significance to us if some tests are implicitly expected to fail. (Not to mention it is a visual cue for a package being maintained or not)

I also skip failures on julia-nightly. They are still visible though if you read the logs.

(I remember there was some discussion of whether 32-bit Windows is actually supported or not..... I forget where that was; maybe a PyCall.jl issue? Or in juliaup?)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,6 +78,8 @@ jobs:
         run: python ci/install_pycall.py
       - name: Run test
         run: python -m tox -- --verbose --cov=julia
+        id: tox-tests
+        continue-on-error: ${{ matrix.julia-version == 'nightly' || (matrix.os == 'windows-latest' && matrix.architecture == 'x86') }}
         env:
           CI: 'true'  # run tests marked by @only_in_ci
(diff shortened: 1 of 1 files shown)


mkitti:
32-bit Windows is not very high on my priority list, so I agree it can be optional. I would like to know what is causing the error though.

MilesCranmer:
Yeah I agree we should try to fix it eventually. (But in the meantime it would be nicer to have it just be an optional check)

MilesCranmer:
I wonder what the other tests keep randomly failing on. Is there any multiprocessing that you know of in the tests? I have had luck preventing segfaults on github actions by specifying a conservative `--heap-size-hint`. Just seems like the garbage collection needs to be more aggressive on action runners otherwise it gets killed.

PR #548, opened 2024-01-01
Add `--bug-report` command line flag

MilesCranmer:
Not sure if adding a unit-test is possible here. I will do local testing in the meantime to ensure this works.


MilesCranmer:
Seems like it doesn't actually get passed to the process for some reason (unlike other options). Opened an issue here: https://github.com/JuliaLang/julia/issues/52689

Issue #555, opened 2024-02-04
Cannot create custom image in Julia 1.10: Method overwriting is not permitted during Module precompilation


denizyuret:
Following the instructions at https://pyjulia.readthedocs.io/en/latest/sysimage.html with Julia 1.10.0 I get the error given in the subject. Here is the full output: https://gist.github.com/denizyuret/dcf47516b0379ce1f7fce8069610d039. The instructions seem to work with Julia 1.9.4.

mkitti:
Do you have `__precompile__(false)` somewhere in your source files?

denizyuret:
I have no source files, just trying to create a plain sys.so that works with conda.

mkitti:
I'm not sure I understand. If you have no new source code, there should not be a reason to create a new system image. What are your objectives? Do you encounter a problem otherwise?

Also have you considered using juliacall? I just deployed a package to here:
https://anaconda.org/julialang/pyjuliacall

```julia
conda install julialang::pyjuliacall
```

denizyuret:
I may be misunderstanding, but the pyjulia documentation says it doesn’t
work by default with conda because the conda python binary is statically
linked (which I confirmed for my setup). Among the several workarounds
suggested is generating a new sys.so image:

https://pyjulia.readthedocs.io/en/latest/troubleshooting.html#your-python-interpreter-is-statically-linked-to-libpython

I will check out pyjuliacall.

mkitti:
I think most settle for the following when using pyjulia.

```python
from julia.api import Julia
jl = Julia(compiled_modules=False)
```

However, I think pyjuliacall probably works better in this situation. Let me know if that's not the case.

mkitti:
If `pyjuliacall` does work better, I would appreciate your feedback on https://github.com/conda-forge/staged-recipes/pull/25097

Issue #542, opened 2023-12-17
Default Conda_env statically linked to libpython?


axla-io:
Hi, I want to be able to use pyjulia, but I run into the following issue:

- I install PyCall
- I open the python executable installed in Conda_env and run julia.install(). This tells me that "PyCall is already installed and compatible with Python executable."
- I run from julia import Base and get an error saying:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 674, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/Users/me/.local/lib/python3.11/site-packages/julia/core.py", line 248, in load_module
    elif self.julia.isafunction(juliapath):
         ^^^^^^^^^^
  File "/Users/me/.local/lib/python3.11/site-packages/julia/core.py", line 239, in julia
    self.__class__.julia = julia = Julia()
                                   ^^^^^^^
  File "/Users/me/.local/lib/python3.11/site-packages/julia/core.py", line 489, in __init__
    raise UnsupportedPythonError(jlinfo)
julia.core.UnsupportedPythonError: It seems your Julia and PyJulia setup are not supported.

Julia executable:
    julia
Python interpreter and libpython used by PyCall.jl:
    /Users/me/.julia/Conda_env/bin/python
    /Users/me/.julia/Conda_env/lib/libpython3.11.dylib
Python interpreter used to import PyJulia and its libpython.
    /Users/me/.julia/Conda_env/bin/python
    /Users/me/.julia/Conda_env/lib/libpython3.11.dylib

Your Python interpreter "/Users/me/.julia/Conda_env/bin/python"
is statically linked to libpython.  Currently, PyJulia does not fully
support such Python interpreter.```

mkitti:
What was the original Python you were trying to use?

mkitti:
Also, have you considered https://github.com/JuliaPy/PythonCall.jl / juliacall ?

PR #556, opened 2024-02-13
Point users to PythonCall.jl

MilesCranmer:
@mkitti What do you think? Currently there is nothing pointing users to PythonCall.jl besides messages in GitHub issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 PyJulia
 =======
 
+> [!CAUTION]
+> Ongoing development of the Python/Julia interface has transitioned to [PythonCall.jl](https://github.com/JuliaPy/PythonCall.jl), please consider using that package instead.
+
 [![Stable documentation](https://img.shields.io/badge/docs-stable-blue.svg)](https://pyjulia.readthedocs.io/en/stable/)
 [![Latest documentation](https://img.shields.io/badge/docs-latest-blue.svg)](https://pyjulia.readthedocs.io/en/latest/)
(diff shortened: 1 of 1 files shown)


mkitti:
I think we should involve a lot more people before proceeding with this.

MilesCranmer:
Fair enough but this has already happened in the various GH issues (eg https://github.com/JuliaPy/pyjulia/issues/555#issuecomment-1925868310, https://github.com/JuliaPy/pyjulia/issues/542#issuecomment-1859135818).

I don’t see a reason it shouldn’t also be mentioned on the README so users don’t accidentally use unmaintained software. Why should they need to post an issue before we tell them to use PythonCall?

jlapeyre:
It seems clear to me that a link the `PythonCall` would be useful.

You could debate about what should be said about the two projects. But just about anything is better than nothing.
The text of this PR looks ok to me.

MilesCranmer:
Seems people are happy with this in https://discourse.julialang.org/t/deprecate-pyjulia-in-favor-of-py-juliacall/111634. Are we good to merge @mkitti? I wouldn't consider this an 'official deprecation' but just a helpful pointer. Maybe I could say 

```diff
- interface has transitioned to PythonCall.jl
+ interface has mostly transitioned to PythonCall.jl
```

to be even more explicit that this is not an official deprecation, if you would like?

jlapeyre:
Just to reiterate and agree with @MilesCranmer . The main points should be easy and not contentious. Add text to landing pages, particularly those with high search visibility that give links to related projects with a bit of context. And this PR hits the main point.

The question of in some sense deprecating pyjulia and anything else that requires any amount of real work can easily be postponed.

Ideally all of the julia/python interface repos should have a section similar to [this one](https://github.com/JeffreySarnoff/ArbNumerics.jl?tab=readme-ov-file#alternatives)

